PR #1, opened 2021-10-21
oct batch - support for encryption key and salt rotation and more

tqinli:
- add .gitignore
- aws enclave build script
- reporting client version
- support for encryption key and salt rotation
- add perf stats per platform
- various fixes and improvements


atarassov-ttd (approved):
LGTM

PR #3, opened 2021-11-01
Fix up the validity time inside refresh token.

atarassov-ttd:
- Fixed the integer overflow when calculating valid till millis
- Change token/refresh endpoint to validate the refresh token is still valid
- Removed deprecated token/refresh-back endpoint


tqinli (approved):
LGTM

PR #4, opened 2021-11-04
Check refresh token expiry based on config (default: false)

atarassov-ttd:
This will allow producing refresh tokens with sensible expiry before enforcing the check. Temporarily configurable for dev purposes: config will be eventually removed.


tqinli (approved):
LGTM

PR #8, opened 2021-12-09
Enforce checking refresh token expiry

atarassov-ttd:
- remove the setting to have this configurable


tqinli (approved):
Thanks for cleaning up the code.

PR #11, opened 2022-02-11
Enable and fix a couple junit vintage tests

haitaottd:
* Update TokenEncodingTest.testRefreshTokenEncoding considering extra valid minute added by encoder
* Add surefire arg to run test using UTF-8 as default charset


atarassov-ttd:
LGTM

PR #13, opened 2022-03-01
Use refresh key to encrypt refresh token. Hide refresh key when responding to key query.

haitaottd:
Use refresh key to encrypt refresh token. Hide refresh key when responding to key query.


atarassov-ttd:
LGTM

PR #16, opened 2022-03-31
UID2-4 create loki logs for collecting referer url

cody-constine-ttd:


src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+
+    private final ConcurrentLinkedQueue<RoutingContext> requestQueue = new ConcurrentLinkedQueue<>();
+    private final Vertx vertx;
+    private ConcurrentHashMap<String, Endpoint> pathMap;
+
+    private static final int MAX_AVAILABLE = 1000;
+    private final Semaphore jsonSemaphore;
+    private final Semaphore queueSemaphore;
+
+    public APIUsageCaptureHandler(long periodicDuration) {
```
atarassov-ttd, line 31: What are the units here?

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+        return a;
+    }
+
+
+    @Override
+    public void handle(RoutingContext routingContext) {
+        routingContext.next();
+        if(!queueSemaphore.tryAcquire()){
+            //Queue is full
+            System.out.println("Queue is full");
```
atarassov-ttd, line 106: This will flood stdout when the queue is full. Would be good to update some micrometer counter though.

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+            return;
+        }
+        requestQueue.offer(routingContext);
+        vertx.runOnContext(this::handleQueue);
```
atarassov-ttd, line 110: Do you need the `requestQueue` at all then? Why not do something like this:

```
vertx.runOnContext(() -> handleRequest(routingContext));
```

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+}
+
+class Domain {
```
atarassov-ttd, line 114: Should this be a nested class to avoid polluting global namespace? The class is not a representation of a domain. Should this be called something like `DomainStats` instead?

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+    private final ConcurrentLinkedQueue<RoutingContext> requestQueue = new ConcurrentLinkedQueue<>();
+    private final Vertx vertx;
+    private ConcurrentHashMap<String, Endpoint> pathMap;
+
+    private static final int MAX_AVAILABLE = 1000;
+    private final Semaphore jsonSemaphore;
+    private final Semaphore queueSemaphore;
+
+    public APIUsageCaptureHandler(long periodicDuration) {
+        vertx = Vertx.vertx();
```
atarassov-ttd, line 32: Starting a whole new vertx and its own thread pool sounds excessive here. Should you be using vetricle(s) and/or perhaps the vertx eventBus instead? This may remove the need for synchronising access to the path map.

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+
+
+public class APIUsageCaptureHandler implements Handler<RoutingContext> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(APIUsageCaptureHandler.class);
+
+    private final ConcurrentLinkedQueue<RoutingContext> requestQueue = new ConcurrentLinkedQueue<>();
+    private final Vertx vertx;
+    private ConcurrentHashMap<String, Endpoint> pathMap;
+
+    private static final int MAX_AVAILABLE = 1000;
```
atarassov-ttd, line 27: You need different constants for max domains to collect and max queue size. Even if their value is the same, the purpose is fundamentally different.

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+
+    @Override
+    public void handle(RoutingContext routingContext) {
+        routingContext.next();
+        if(!queueSemaphore.tryAcquire()){
+            //Queue is full
+            System.out.println("Queue is full");
+            return;
+        }
+        requestQueue.offer(routingContext);
```
atarassov-ttd, line 109: I don't think it is a good idea to queue the whole routing context object. I wonder what impact this might have. Perhaps easier to extract the necessary items and bundle them in a `QueueItem` object.

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+    public void handleQueue(Void v) {
+        RoutingContext routingContext = requestQueue.poll();
+        queueSemaphore.release();
+        assert routingContext != null;
+        String path = routingContext.request().path();
+        String apiVersion = "v0";
+        String endpoint = path.substring(1);
+
+        if(path.charAt(1) == 'v') {
+            int apiVIndex = path.indexOf("/", 1);
```
atarassov-ttd, line 72: What happens if this returns -1?

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+            apiVersion = path.substring(1, apiVIndex);
+            endpoint = path.substring(apiVIndex+1);
+        }
+
+        String referer = routingContext.request().headers().get("Referer");
```
atarassov-ttd, line 77: What happens if there is no such header?
Shouldn't this parse the domain name out of the referer url?

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+        ClientKey clientKey = (ClientKey) AuthMiddleware.getAuthClient(routingContext);
+        String apiContact = clientKey.getContact();
+
+        Integer siteId = clientKey.getSiteId();
+        Domain domain = new Domain(referer, 1, apiContact);
+
+        Endpoint endpointClass = new Endpoint(endpoint, siteId, apiVersion, domain);
+
+        try {
+            jsonSemaphore.acquire();
```
atarassov-ttd, line 87: What is this for?

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    public void handleJsonSerial(Long l) {
+        try {
+            jsonSemaphore.acquire(MAX_AVAILABLE);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+
+        Object[] localValues = pathMap.values().toArray();
```
atarassov-ttd, line 46: Maybe worth using an AtomicReference for `pathMap` and just switch over without having to hold the lock.

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,205 @@
...
+        vertx = Vertx.vertx();
+        pathMap = new ConcurrentHashMap<>();
+        vertx.setPeriodic(periodicDuration, this::handleJsonSerial);
+        jsonSemaphore = new Semaphore(MAX_AVAILABLE, true);
+        queueSemaphore = new Semaphore(MAX_AVAILABLE, true);
+    }
+
+    public void handleJsonSerial(Long l) {
+        try {
+            jsonSemaphore.acquire(MAX_AVAILABLE);
```
atarassov-ttd, line 41: Looks like you are using semaphore as mutex. Do you need locking at all? If so, why not use plain synchronized blocks?

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
cody-constine-ttd: Then its a v0 api

src/main/java/com/uid2/operator/vertx/APIUsageCaptureHandler.java
cody-constine-ttd: I was actually using it as a reverse reader writer, because we can allow many writers at once thanks to .merge() but I wanted to stop all of them to capture the state and clear it.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -153,6 +153,7 @@ private Router createRoutesSetup() throws IOException {
             .allowedHeader("Content-Type"));
         router.route().handler(BodyHandler.create().setBodyLimit(MAX_REQUEST_BODY_SIZE));
 
+        router.route().handler(new APIUsageCaptureHandler(60000));
```
haitaottd, line 156: Do all requests have referer header? I thought only request from browser has that. Only refresh request is from browser.

cody-constine-ttd: We are anticipating that requests will not have referer url

src/main/java/com/uid2/operator/monitoring/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,303 @@
...
+                .register(Metrics.globalRegistry);
+    }
+
+    public void handleMessage(Message message) {
+        ObjectMapper mapper = new ObjectMapper();
+        MessageItem messageItem = null;
+        try {
+            messageItem = mapper.readValue(message.body().toString(), MessageItem.class);
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
```
tqinli, line 76: log through logger instead of print?

src/main/java/com/uid2/operator/monitoring/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,303 @@
...
+        if(!queueSemaphore.tryAcquire()){
+            queueFullCounter.increment();
+            return;
+        }
+
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            vertx.eventBus().send("APIUsage", mapper.writeValueAsString(messageItem));
+        } catch (JsonProcessingException e) {
+            e.printStackTrace();
```
tqinli, line 136: same here, switch to logger?

src/main/java/com/uid2/operator/monitoring/APIUsageCaptureHandler.java
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        String referer = messageItem.getReferer();
+        if(referer == null){
+            referer = "unknown";
+        } else {
+            try {
+                referer = new URI(referer).getHost();
+            } catch (URISyntaxException e) {
+                e.printStackTrace();
```
tqinli, line 98: same here, switch to logger

tqinli (approved):
LGTM

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,255 @@
...
+            return domainList;
+        }
+
+        public void Merge(EndpointStat other) {
+            other.domainList.forEach(this::AddDomain);
+        }
+
+        public void AddDomain(DomainStat d) {
+            String domainName = d.getDomain();
+            if(domainMap.containsKey(domainName)) {
```
atarassov-ttd, line 231: Is there a way/need to avoid a double lookup here?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,255 @@
...
+
+    class EndpointStat {
+        private String endpoint;
+        private Integer siteId;
+        private String apiVersion;
+        private ArrayList<DomainStat> domainList;
+        private HashMap<String, Integer> domainMap;
+
+
+        private final int DomainSize = 1000;
```
atarassov-ttd, line 197: How about `MaxDomains` or something along those lines?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,255 @@
...
+            else if(domainList.size() < DomainSize) {
+                domainList.add(d);
+                domainMap.put(domainName, domainList.size()-1);
+            } else {
+                domainMissedCounter.increment();
+            }
+        }
+
+        public String toString() {
+            StringBuilder outString = new StringBuilder("endpoint: " + endpoint + ", site_id: " + siteId + ", api_version: " + apiVersion +
```
atarassov-ttd, line 243: A bit weird to see string builder initialised with a bunch of string concatenations. Would using JSON all the way be a good option here?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,255 @@
...
+                    "\nDomains: ");
+
+            for (int i = 0; i < domainList.size(); i++) {
```
atarassov-ttd, line 246: nit: `for (DomainStat domainStat : domainList)`? Might also want to consider using streams.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -598,6 +608,33 @@ private void recordIdentityMapStats(RoutingContext rc, int inputCount) {
         ds.record(inputCount);
     }
 
+    public void sendToStatsCollector(RoutingContext routingContext) {
```
atarassov-ttd, line 611: Could you please extract this logic into its own class? The operator verticle is already bloated enough.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -598,6 +608,33 @@ private void recordIdentityMapStats(RoutingContext rc, int inputCount) {
...
+        routingContext.next();
+        assert routingContext != null;
+
+        String path = routingContext.request().path();
+        String referer = routingContext.request().headers().get("Referer");
+        ClientKey clientKey = (ClientKey) AuthMiddleware.getAuthClient(routingContext);
+        StatsCollectorMessageItem messageItem = new StatsCollectorMessageItem(path, referer, clientKey.getContact(), clientKey.getSiteId());
+
+        if(_statCollectorCount.get() >= MAX_STAT_COLLECTORS){
+            Counter queueFullCounter = Counter
```
atarassov-ttd, line 621: Should this be a class member?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -598,6 +608,33 @@ private void recordIdentityMapStats(RoutingContext rc, int inputCount) {
...
+                    .builder("uid2.api_usage_queue_full")
+                    .description("counter for how many usage messages are dropped because the queue is full")
+                    .register(Metrics.globalRegistry);
+            queueFullCounter.increment();
+            return;
+        }
+
+        ObjectMapper mapper = new ObjectMapper();
+        try {
+            vertx.eventBus().send("StatsCollector", mapper.writeValueAsString(messageItem));
```
atarassov-ttd, line 631: The `StatsCollector` string is repeated here and in the recipient verticle. Would be great to make this is a const.

In general though, even better would be to wrap the logic of sending the message and doing the queue size check behind an interface and have `StatsCollectorVerticle` implement that interface (since it own the queue).

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -976,4 +978,40 @@ private void generateRefreshToken(Vertx vertx, String email, int siteId, Handler
...
+        });
+    }
+    @Test void sendToStatsCollectorFullQueue(Vertx vertx, VertxTestContext testContext) {
+        final int clientSiteId = 201;
+        final String emailAddress = "test@uid2.com";
+        fakeAuth(clientSiteId, Role.GENERATOR);
+        setupSalts();
+        setupKeys();
+
+        statsCollectorRunning.set(1000);
```
atarassov-ttd, line 1006: Where does this 1000 come from?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -976,4 +978,40 @@ private void generateRefreshToken(Vertx vertx, String email, int siteId, Handler
...
+    @Test void sendInformationToStatsCollector(Vertx vertx, VertxTestContext testContext) {
+        final int clientSiteId = 201;
+        final String emailAddress = "test@uid2.com";
+        fakeAuth(clientSiteId, Role.GENERATOR);
+        setupSalts();
+        setupKeys();
+
+        vertx.eventBus().consumer("StatsCollector", message -> {
+            String expected = "{\"path\":\"/v1/token/generate\",\"referer\":null,\"apiContact\":null,\"siteId\":201}";
+            assert message.body().toString().equals(expected);
```
atarassov-ttd, line 991: Have you considered using junit/jupiter assertion methods?

src/test/java/com/uid2/operator/StatsCollectorVerticleTest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Test
+    void atomicIntDecremented(Vertx vertx, VertxTestContext testContext) throws JsonProcessingException, InterruptedException {
+        statsCollectorRunning.incrementAndGet();
+
+        ObjectMapper mapper = new ObjectMapper();
+        StatsCollectorMessageItem messageItem = new StatsCollectorMessageItem("test", "https://test.com", "test", 1);
+
+        vertx.eventBus().send("StatsCollector", mapper.writeValueAsString(messageItem));
+        testContext.awaitCompletion(1000, TimeUnit.MILLISECONDS);
```
atarassov-ttd, line 67: Can this timing logic be replaced by using a mock clock? Would it make sense to add some levels of abstraction to the stats collector verticle implementation to enable you mock the timing/thread stuff out?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,255 @@
+package com.uid2.operator.monitoring;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
```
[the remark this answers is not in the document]
cody-constine-ttd: JSON is vert.x recommendation. Do you have another way?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,255 @@
...
+        this.jsonSerializerExecutor = vertx.createSharedWorkerExecutor("stats-collector-json-worker-pool");
+    }
+
+    public void handleMessage(Message message) {
+        ObjectMapper mapper = new ObjectMapper();
+        StatsCollectorMessageItem messageItem = null;
+        try {
+            messageItem = mapper.readValue(message.body().toString(), StatsCollectorMessageItem.class);
+        } catch (JsonProcessingException e) {
+            LOGGER.error(e.getMessage(), e);
```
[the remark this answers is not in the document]
cody-constine-ttd: I don't think it will ever fail, but java requires to catch the exception.

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,255 @@
...
+        _statsCollectorCount.decrementAndGet();
+
+        if(Duration.between(lastJsonProcessTime, Instant.now()).toMillis() >= jsonProcessingInterval){
+            lastJsonProcessTime = Instant.now();
+            if(_runningSerializer > 0){
+               logCycleSkipperCounter.increment();
+            } else {
+                _runningSerializer ++;
+                this.jsonSerializerExecutor.<String>executeBlocking(
+                        promise -> promise.complete(this.serializeToLogs(pathMap.values().toArray())),
```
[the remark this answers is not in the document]
cody-constine-ttd: image

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
cody-constine-ttd: The max number of stats collectors?

src/test/java/com/uid2/operator/StatsCollectorVerticleTest.java
cody-constine-ttd: No... I tried the vertx junit framework is weird. It extracts the tests in a strange way, so this is the only way to get things done.

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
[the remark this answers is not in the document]
cody-constine-ttd: Ok it was not thread safe, but it was because the promise gets executed after the rest of the thread too. So I fixed it, by capturing the array before the promise.

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
cody-constine-ttd: Is this an issue its a just a hashmap lookup?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,250 @@
...
+            return;
+        }
+
+        assert messageItem != null;
+
+        String path = messageItem.getPath();
+        String apiVersion = "v0";
+        String endpoint = path.substring(1);
+
+        if(path.charAt(1) == 'v') {
```
atarassov-ttd, line 85: Can `path` be null? What would happen if this or anything below throws?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,250 @@
...
+        }
+
+        String referer = messageItem.getReferer();
+        if(referer == null){
+            referer = "unknown";
+        } else {
+            try {
+                referer = new URI(referer).getHost();
+            } catch (URISyntaxException e) {
+                LOGGER.error(e.getMessage(), e);
```
atarassov-ttd, line 98: Given referer string comes from client, I think we should avoid logging a message on error.

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,250 @@
...
+            if(_runningSerializer){
+               logCycleSkipperCounter.increment();
+            } else {
+                _runningSerializer = true;
+                Object[] stats = pathMap.values().toArray();
+                this.jsonSerializerExecutor.<Void>executeBlocking(
+                        promise -> promise.complete(this.serializeToLogs(stats)),
+                        res -> {
+                            if(!res.succeeded()) {
+                                LOGGER.error("Failed To Serialize JSON");
```
atarassov-ttd, line 123: Also log error details?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,250 @@
...
+        return null;
+    }
+
+    private EndpointStat mergeEndpoint(EndpointStat a, EndpointStat b) {
+        a.merge(b);
+        return a;
+    }
+
+
+    public String GetEndpointStats() {
```
atarassov-ttd, line 154: `getEndpointStats`?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,250 @@
...
+                completeStats.append(jsonString).append("\n");
+            } catch (JsonProcessingException e) {
+                LOGGER.error(e.getMessage(), e);
+            }
+        }
+        return completeStats.toString();
+    }
+
+    static class DomainStat {
+        private String domain;
```
atarassov-ttd, line 169: `domain` and `apiContact` can be `final`?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,250 @@
...
+            return apiContact;
+        }
+
+        public void merge(DomainStat d) {
+            count += d.getCount();
+        }
+    }
+
+    class EndpointStat {
+        private String endpoint;
```
atarassov-ttd, line 197: Some of these fields can be `final`?

src/test/java/com/uid2/operator/StatsCollectorVerticleTest.java
```diff
@@ -0,0 +1,99 @@
...
+        vertx.eventBus().send(Const.Config.StatsCollectorEventBus, mapper.writeValueAsString(messageItem));
+        vertx.eventBus().send(Const.Config.StatsCollectorEventBus, mapper.writeValueAsString(messageItem));
+        vertx.eventBus().send(Const.Config.StatsCollectorEventBus, mapper.writeValueAsString(messageItem));
+
+        messageItem = new StatsCollectorMessageItem("/v1/test", "https://test.com", "test", 1);
+
+        vertx.eventBus().send(Const.Config.StatsCollectorEventBus, mapper.writeValueAsString(messageItem));
+        vertx.eventBus().send(Const.Config.StatsCollectorEventBus, mapper.writeValueAsString(messageItem));
+
+        testContext.awaitCompletion(2000, TimeUnit.MILLISECONDS);
```
atarassov-ttd, line 87: Is it possible to have these tests not depend on clock?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
cody-constine-ttd: If it gets here the called function will have thrown an exception.

src/test/java/com/uid2/operator/StatsCollectorVerticleTest.java
cody-constine-ttd: I have tried all the ways I can't think of but Vert.x is not being friendly

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
cody-constine-ttd: It should never be null its only set by our code. If anything throws in here, we will lose the single message.

src/main/java/com/uid2/operator/monitoring/StatsCollectorHandler.java
```diff
@@ -0,0 +1,46 @@
...
+import com.uid2.shared.middleware.AuthMiddleware;
+import io.micrometer.core.instrument.Counter;
+import io.micrometer.core.instrument.Metrics;
+import io.vertx.core.Handler;
+import io.vertx.core.Vertx;
+import io.vertx.ext.web.RoutingContext;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.atomic.AtomicInteger;
```
atarassov-ttd, line 18: unused?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,271 @@
...
+                .description("counter for how many usage messages are dropped because the queue is full")
+                .register(Metrics.globalRegistry);
+
+        mapper = new ObjectMapper();
+    }
+
+    @Override
+    public void start() throws Exception {
+        super.start();
+        vertx.eventBus().consumer(Const.Config.StatsCollectorEventBus, this::handleMessage);
```
atarassov-ttd, line 72: nit: start consuming messages after you've initialised all the necessary things (i.e. this should go to the end of the function)

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,271 @@
...
+            return;
+        }
+
+        assert messageItem != null;
+
+        String path = messageItem.getPath();
+        String apiVersion = "v0";
+        String endpoint = path.substring(1);
+
+        if(path.charAt(1) == 'v') {
```
atarassov-ttd, line 92: Should this be: `if (path.length() > 1 && path.charAt(1) == 'v')`?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,271 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+
+
+public class StatsCollectorVerticle extends AbstractVerticle implements IStatsCollectorQueue {
+    private static final Logger LOGGER = LoggerFactory.getLogger(StatsCollectorVerticle.class);
+    private HashMap<String, EndpointStat> pathMap;
+
+    private static final int MAX_AVAILABLE = 1000;
+
+    private final long jsonProcessingInterval;
```
atarassov-ttd, line 30: nit: turn this into a `Duration`

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -0,0 +1,271 @@
...
+                            _runningSerializer = false;
+                        }
+                );
+                pathMap.clear();
+            }
+
+        }
+    }
+
+    Void serializeToLogs(Object[] stats) {
```
atarassov-ttd, line 140: `private void`?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -47,7 +47,7 @@ public StatsCollectorVerticle(long jsonIntervalMS) {
         _statsCollectorCount = new AtomicInteger();
         _runningSerializer = false;
 
-        jsonProcessingInterval = jsonIntervalMS;
+        jsonProcessingInterval = Duration.ofMillis(jsonIntervalMS-1);
```
atarassov-ttd, line 50: Why -1?

src/main/java/com/uid2/operator/monitoring/StatsCollectorVerticle.java
```diff
@@ -69,9 +69,10 @@ public StatsCollectorVerticle(long jsonIntervalMS) {
     @Override
     public void start() throws Exception {
         super.start();
-        vertx.eventBus().consumer(Const.Config.StatsCollectorEventBus, this::handleMessage);
         this.jsonSerializerExecutor = vertx.createSharedWorkerExecutor("stats-collector-json-worker-pool");
-        lastJsonProcessTime = Instant.ofEpochMilli(Instant.now().toEpochMilli() + jsonProcessingInterval - 1);
+        //lastJsonProcessTime = Instant.ofEpochMilli(Instant.now().toEpochMilli() + jsonProcessingInterval - 1);
```
atarassov-ttd, line 73: dead code?

atarassov-ttd (approved):

PR #19, opened 2022-04-06
Add AES-GCM encryption and decryption methods

haitaottd:


tqinli (approved):
LGTM

src/main/java/com/uid2/operator/service/EncryptionHelper.java
```diff
@@ -91,7 +96,49 @@ public static byte[] decrypt(byte[] encryptedBytes, SecretKey key) {
...
+
+    public static byte[] encryptGCM(byte[] b, byte[] secretBytes) {
+        try {
+            final SecretKey k = new SecretKeySpec(secretBytes, "AES");
+            final Cipher c = Cipher.getInstance(gcmCipherScheme);
+            final byte[] ivBytes = new byte[16];
+            threadLocalSecureRandom.get().nextBytes(ivBytes);
+            GCMParameterSpec gcmParameterSpec = new GCMParameterSpec(GCM_TAG_LENGTH * 8, ivBytes);
+            c.init(Cipher.ENCRYPT_MODE, k, gcmParameterSpec);
+            return Buffer.buffer().appendBytes(ivBytes).appendBytes(c.doFinal(b)).getBytes();
```
atarassov-ttd, line 118: Where does the GCM auth tag go?

src/main/java/com/uid2/operator/service/V2EncryptedTokenEncoder.java
```diff
@@ -37,8 +37,11 @@ public class V2EncryptedTokenEncoder implements ITokenEncoder {
     private final IKeyStore keyStore;
     private final SecureRandom random = new SecureRandom();
 
-    public V2EncryptedTokenEncoder(IKeyStore keyStore) {
+    private final boolean useGCM;
+
+    public V2EncryptedTokenEncoder(IKeyStore keyStore, boolean useGCM) {
```
atarassov-ttd, line 42: Should this be a new class `V3EncryptedTokenEncoder` instead? GCM and CBC encrypted tokens are not interoperable, the format has to indicate which encryption is used. Thus I think it is better to introduce the new formatter.

haitaottd: I don't think so. Token is just encrypted differently. The unencrypted format doesn't change.

atarassov-ttd: Unencryption is done by DSPs using DSP SDK. How would the SDK know looking at a blob which algorithm to use to decrypt the content?

haitaottd: They can tell by token length. Advertising token is 180 bytes with CBC, 184 bytes with GCM

PR #20, opened 2022-04-06
Uid2 88 aws read config from secret manager

scong-ttd:
Load secret/config from secrets manager instead of user data.

New pipe (inside-out):
- enclave /etc/paths: 127.0.0.1 secretsmanager.$REGION.amazonaws.com
- enclave vsockpx: 127.0.0.1:443 -> vsock 3:3308
- host https-proxy: vsock -1:3308 -> secretsmanager.$REGION.amazonaws.com 443


tqinli (approved):
LGTM. Thanks!

PR #23, opened 2022-05-02
Refactor to support id and token format v3

atarassov-ttd:
* Make data model simpler and more explicit about intentions
* Remove dead/legacy code


src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
```diff
@@ -29,114 +29,132 @@
...
+            : encodeV2(t, masterKey, siteEncryptionKey);
+    }
 
-        // <version><space><master key id><space>
+    private byte[] encodeV2(AdvertisingToken t, EncryptionKey masterKey, EncryptionKey siteKey) {
+        final Buffer b = Buffer.buffer();
 
-        b.appendByte((byte) t.getVersion());
-        b.appendByte(FIELD_SEPARATOR);
+        b.appendByte((byte) t.version.rawVersion);
```
tqinli, line 55: no field separate for encodeV2, is that right or?

src/main/java/com/uid2/operator/service/InputUtil.java
```diff
@@ -130,8 +136,8 @@ public static String NormalizeEmailString(String email) {
         return addressPartToUse.append('@').append(domainPart).toString();
     }
 
-    public static enum Type {
-        Email,
+    public enum IdentityInputType {
+        Raw,
```
tqinli, line 140: maybe call this RawEmail? Or maybe leaving it as Email is ok?

tqinli (approved):
LGTM

PR #27, opened 2022-05-09
IDv3 support for generating and parsing tokens and UID string

atarassov-ttd:
- config controls whether ID v2 or v3 is to be generated for UID2 emails
- config controls whether v2 or v3 advertising/refresh tokens are to be generated
- enable parsing both v2 and v3 tokens

### DIFF
--- a/src/main/java/com/uid2/operator/service/UIDOperatorService.java
+++ b/src/main/java/com/uid2/operator/service/UIDOperatorService.java
@@ -57,6 +57,8 @@ public class UIDOperatorService implements IUIDOperatorService {
     private final Duration refreshIdentityAfter;
 
     private final OperatorIdentity operatorIdentity;
+    private final TokenVersion tokenVersion;
+    private final boolean uid2EmailV3Enabled;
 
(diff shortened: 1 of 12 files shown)


src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -81,6 +83,9 @@ public UIDOperatorService(JsonObject config, IOptOutStore optOutStore, ISaltProv
         if (this.refreshIdentityAfter.compareTo(this.refreshExpiresAfter) > 0) {
             throw new IllegalStateException(REFRESH_TOKEN_EXPIRES_AFTER_SECONDS + " must be >= " + REFRESH_IDENTITY_TOKEN_AFTER_SECONDS);
         }
+
+        this.tokenVersion = config.getBoolean("generate_v3_tokens", false) ? TokenVersion.V3 : TokenVersion.V2;
+        this.uid2EmailV3Enabled = config.getBoolean("uid2_email_v3", false);
```
haitaottd, line 88: why two v2 flags?

tqinli: this is to allow phone support to switch to v2 token independently later, I think?

tqinli (approved):
LGTM

PR #32, opened 2022-05-17
Support phone/phonehash in V2 API

haitaottd:


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -687,15 +701,35 @@ private InputUtil.InputVal getTokenInputV2(JsonObject req) {
         if (req == null)
             return null;
 
+        Supplier<InputUtil.InputVal> getInput = null;
```
tqinli, line 704: feel like there is a way to implement without lambda?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -392,7 +392,7 @@ private void setupSiteKey(int siteId, int keyId) {
     }
 
     private void generateTokens(String apiVersion, Vertx vertx, String inputType, String input, Handler<JsonObject> handler) {
-        String v1Param = inputType + "=" + input;
+        String v1Param = inputType + "=" + urlEncode(input);
```
tqinli, line 395: I think there are existing tests that calls urlEncode for input before calling this. Maybe revert those if you haven't?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -857,22 +891,50 @@ private void handleIdentityMapBatchV1(RoutingContext rc) {
...
+                    ResponseUtil.ClientError(rc, "Exactly one of [email, email_hash, phone, phone_hash] must be specified");
+                    return;
+                }
+                getInputList = () -> createInputListV1(phones, IdentityType.Phone, InputUtil.IdentityInputType.Raw);
+            }
+
+            final JsonArray phoneHashes = obj.getJsonArray("phone_hash");
+            if (phoneHashes != null && !phoneHashes.isEmpty()) {
+                if (getInputList != null) {
+                    ResponseUtil.ClientError(rc, "Exactly one of [email, email_hash, phone, phone_hash] must be specified");
```
atarassov-ttd, line 923: This keeps getting repeated. Can the code be reorganised a bit to avoid the repetition?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
haitaottd: All reverted.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
haitaottd: hmm, can't think of another way. Using lambda so that I don't need to recheck which one of email/email_hash/phone/phone_hash is valid input at the end.

atarassov-ttd:
LGTM

PR #36, opened 2022-06-01
Allow setupGoogleTag in JS SDK to be called prior to googleTag loading

TLadd:
Presently, if `googletag` is not defined on the window when the uid2 js script is loaded, the `encryptedSignalProviders.push` call is not made and therefore the `collectorFunction` is never called.

In this situation where `googletag` is not defined, you can define`window.googletag.encryptedSignalProviders` with the desired object and googletag will pick it up when its js loads. This PR implements that change.

### DIFF
--- a/static/js/uid2-sdk-1.0.0.js
+++ b/static/js/uid2-sdk-1.0.0.js
@@ -34,18 +34,22 @@ class UID2 {
     }
 
     static setupGoogleTag() {
-      if (window.googletag && window.googletag.encryptedSignalProviders) {
-          googletag.encryptedSignalProviders.push({
-              id: 'uidapi.com',
-              collectorFunction: () => {
-                  if (window.__uid2 && window.__uid2.getAdvertisingToken) {
-                      return __uid2.getAdvertisingTokenAsync();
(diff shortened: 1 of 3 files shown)


static/js/uid2-sdk-1.0.0.js
```diff
@@ -34,18 +34,22 @@ class UID2 {
...
+        if (!window.googletag) {
+            window.googletag = {};
+        }
+        if (!googletag.encryptedSignalProviders) {
+            googletag.encryptedSignalProviders = [];
+        }
+        googletag.encryptedSignalProviders.push({
+            id: "uidapi.com",
+            collectorFunction: () => {
+                if (window.__uid2 && window.__uid2.getAdvertisingTokenAsync) {
```
TLadd, line 46: I changed this check in the collectorFunction from `getAdvertisingToken` to `getAdvertisingTokenAsync` since `getAdvertisingTokenAsync` is the method used within.

cody-constine-ttd (approved):

atarassov-ttd (approved):

PR #52, opened 2022-07-25
Static has to be before the stats collector?

cody-constine-ttd:


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -178,13 +178,13 @@ private Router createRoutesSetup() throws IOException {
             .allowedHeader("Content-Type"));
         router.route().handler(BodyHandler.create().setBodyLimit(MAX_REQUEST_BODY_SIZE));
 
+        router.route("/static/*").handler(StaticHandler.create("static"));
         router.route().handler(new StatsCollectorHandler(_statsCollectorQueue, vertx));
```
atarassov-ttd, line 182: Should this happen after setting up v2 routes?

cody-constine-ttd: It has to happen before. Routes are executed in order and the v2 routes terminate the context

PR #53, opened 2022-07-26
Update SDK callback parameter naming

dlackty:
According to UnifiedID2/uid2docs#42's feedback from @elenaTTD, I updated the callback naming to make it consistent between doc and real code behavior.

SDK v1 actually has same inconsistency between doc and codes. Should v1's SDK code be renamed too? 

Thank you.


atarassov-ttd:
Hi @dlackty and thank you for the PR! Can you please make the following amendments to the proposed changes?

1. Rather than updating the field, could you instead add a new field with the correct name? Updating field name is a breaking change that we would like to avoid.
2. As you mentioned, v1 SDK has the same problem. Do you mind updating that as well as the EUID SDK too?

Thanks!

dlackty:
@ atarassov-ttd I updated the PR according to your suggestions. Please have a look. Thank you!

atarassov-ttd (approved):
LGTM

PR #57, opened 2022-08-30
EUID marketplace offering

cody-constine-ttd:


scripts/aws/Dockerfile
```diff
@@ -4,9 +4,11 @@ WORKDIR /app
 ARG JAR_NAME=uid2-operator
 ARG JAR_VERSION=1.0.0
 ARG IMAGE_VERSION=1.0.0.unknownhash
+ARG REGION=UID2
```
atarassov-ttd, line 7: Should this be renamed to `SCOPE` or `IDENTITY_SCOPE` to match the setting name in the config file?

scripts/aws/config.json
```diff
@@ -1 +1,14 @@
-{ "service_instances": 4, "clients_metadata_path": "https://core-prod.uidapi.com/clients/refresh", "keys_metadata_path": "https://core-prod.uidapi.com/key/refresh", "keys_acl_metadata_path": "https://core-prod.uidapi.com/key/acl/refresh", "salts_metadata_path": "https://core-prod.uidapi.com/salt/refresh", "optout_metadata_path": "https://optout-prod.uidapi.com/optout/refresh", "core_attest_url": "https://core-prod.uidapi.com/attest", "core_api_token": "your-api-key", "optout_s3_path_compat": true, "optout_api_uri": "https://core-prod.uidapi.com/optout/replicate", "optout_api_token": "your-api-key", "enclave_platform": "aws-nitro" }
```
atarassov-ttd, line 1: What is this file used for?

scripts/aws/conf/integ-config.json
```diff
@@ -0,0 +1,10 @@
+{
```
atarassov-ttd, line 1: Should this file be named `integ-uid2-config.json` for consistency?

Makefile.nitro
```diff
@@ -45,9 +49,20 @@ build/load_config.py: ./scripts/aws/load_config.py
 build/make_config.py: ./scripts/aws/make_config.py
 	cp ./scripts/aws/make_config.py ./build/
 
+build/configs: build/conf/prod-config.json build/conf/integ-config.json build/conf/prod-euid-config.json build/conf/integ-euid-config.json
+
 build/conf/prod-config.json: build_artifacts ./scripts/aws/conf/prod-config.json
```
atarassov-ttd, line 54: Should this file be renamed to `prod-uid2-config.json` for consistency?

scripts/aws/make_config.py
```diff
@@ -14,8 +14,9 @@ def apply_override(config, overrides, key, type):
 
 
 config_path = sys.argv[1]
-overrides_path = sys.argv[2]
-thread_count = int(sys.argv[3])
+integ_config_path = sys.argv[2]
+overrides_path = sys.argv[3]
+thread_count = int(sys.argv[4])
 
 config = load_json(config_path)
```
atarassov-ttd, line 21: Rather than have overrides section for integ below, have you considered doing something like this:

```
overrides = load_json(overrides_path)
config = load_json(integ_config_path if override.get('environment') == 'integ' else prod_config_path)
```

scripts/aws/config.json
cody-constine-ttd: I think was just used for testing

scripts/aws/make_config.py
cody-constine-ttd: I could be convinced but this way makes all the other settings besides the URLs easier to keep in sync

scripts/aws/EUID_CloudFormation.template.yml
```diff
@@ -0,0 +1,433 @@
...
+          - VpcSubnet1
+          - VpcSubnet2
+          - NewVpcCidr
+          - Subnet1Cidr
+          - Subnet2Cidr
+    ParameterLabels:
+      APIToken:
+        default: OPERATOR_KEY provided by EUID Administrator
+      DeployToEnvironment:
+        default: EUID environment to deploy to. Prod - production; Integ - integration test.
```
atarassov-ttd, line 116: Should `prod` and `integ` be lowercase here?

scripts/aws/EUID_CloudFormation.template.yml
```diff
@@ -0,0 +1,433 @@
...
+        default: If choose to create new VPC, enter CIDR for new VPC (required in case of new VPC)
+      CustomizeEnclaceResource:
+        default: Enclave resouce configuration auto calculated or manual
+      EnclavememoryinMB:
+        default: If choose to false for CustomizeEnclaceResource, enter memory for Enclave in MB
+      EnclaveCPUCount:
+        default: If choose to false for CustomizeEnclaceResource, enter CPU count for Enclave
+Mappings:
+  RegionMap:
+    us-east-1:
```
atarassov-ttd, line 141: Should this list be reduced to only EU locations?

scripts/aws/Dockerfile
```diff
@@ -4,9 +4,11 @@ WORKDIR /app
 ARG JAR_NAME=uid2-operator
 ARG JAR_VERSION=1.0.0
 ARG IMAGE_VERSION=1.0.0.unknownhash
+ARG IDENTITY_SCOPE=UID2
```
atarassov-ttd, line 7: Should the default value be removed here?

scripts/aws/entrypoint.sh
```diff
@@ -24,7 +24,12 @@ echo "AWS_REGION_NAME=$AWS_REGION_NAME"
 echo "127.0.0.1 secretsmanager.$AWS_REGION_NAME.amazonaws.com" >> /etc/hosts
 
 python3 /app/load_config.py >/app/conf/config-overrides.json
-python3 /app/make_config.py /app/conf/config.json /app/conf/config-overrides.json $(nproc) >/app/conf/config-final.json
+
+if [ "$IDENTITY_SCOPE" = 'UID2' ]; then
+  python3 /app/make_config.py /app/conf/prod-uid2-config.json /app/conf/integ-uid2-config.json /app/conf/config-overrides.json $(nproc) >/app/conf/config-final.json
+elif [ "$IDENTITY_SCOPE" = 'EUID' ]; then
+  python3 /app/make_config.py /app/conf/prod-euid-config.json /app/conf/integ-euid-config.json /app/conf/config-overrides.json $(nproc) >/app/conf/config-final.json
+fi
```
atarassov-ttd, line 32: Add `else` block and explicitly fail with an error message?

scripts/aws/entrypoint.sh
```diff
@@ -11,7 +11,7 @@ ifconfig lo 127.0.0.1
 /app/vsockpx --config /app/proxies.nitro.yaml --daemon --workers $(( $(nproc) * 2 )) --log-level 3
 
 # -- load config via proxy
-export UID2_CONFIG_SECRET_KEY=$([[ "$(curl -s -x socks5h://127.0.0.1:3305 http://169.254.169.254/latest/user-data | grep UID2_CONFIG_SECRET_KEY=)" =~ ^export\ UID2_CONFIG_SECRET_KEY=\"(.*)\"$ ]] && echo ${BASH_REMATCH[1]} || echo "uid2-operator-config-key")
+export UID2_CONFIG_SECRET_KEY=$([[ "$(curl -s -x socks5h://127.0.0.1:3305 http://169.254.169.254/latest/user-data | grep UID2_CONFIG_SECRET_KEY=)" =~ ^export\ UID2_CONFIG_SECRET_KEY=\"(.*)\" ]] && echo ${BASH_REMATCH[1]} || echo "uid2-operator-config-key")
```
atarassov-ttd, line 14: Why is this changed?

scripts/aws/make_config.py
atarassov-ttd: But then you still have two files with full set of configs which is very confusing. I see two options here:
1. Have prod config to contain all settings and have integ config only have the (overridden) paths.
2. Have a single base config without paths and then have prod/integ config only with paths.

I like the second approach more since in the first one if a new path is added to prod config, but not to integ, the operator would behave weirdly which would be more difficult to troubleshoot.

scripts/aws/entrypoint.sh
cody-constine-ttd: The regex wasn't working with the `$` at the end

scripts/aws/Dockerfile
cody-constine-ttd: It feels like UID2 should be the default value.

scripts/aws/config.json
scong-ttd: This used to be the config template that enclave consumes. Doesn't seem to be used anymore.

Makefile.nitro
```diff
@@ -33,9 +33,13 @@ clean:
 	cd build/dante-1.4.3; ./configure
 	cd build/dante-1.4.3; make
 
-build_eif: uid2operator.eif
+build_eif: uid2operator.eif euidoperator.eif
```
scong-ttd, line 36: Do you plan to send this euidoperator.eif onto the AMI and rename it to be uid2operator? Otherwise we need to `uid2operator.service`

scripts/aws/entrypoint.sh
scong-ttd, line 14: This is not consuming `EUID_CONFIG_SECRET_KEY`

PR #75, opened 2022-12-06
UID2-532 Update-operator-image

cYKatherine:
Update openjdk11 to eclipse-termurin11


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM openjdk:11
+FROM eclipse-temurin:11.0.17_8-jre-alpine
```
scong-ttd, line 1: Shall we add sha256 digest?

PR #80, opened 2023-01-13
Using new version of Shared that introduces scoping stores to either …

aulme:
…site specific or global context. For now only using global scope.


pom.xml
```diff
@@ -19,7 +19,7 @@
         <!-- check micrometer.version vertx-micrometer-metrics consumes before bumping up -->
         <micrometer.version>1.1.0</micrometer.version>
         <enclave-api.version>1.1.0</enclave-api.version>
-        <uid2-shared.version>1.12.0</uid2-shared.version>
+        <uid2-shared.version>1.16.0</uid2-shared.version>
```
sunnywu, line 22: waiting for shawn to confirm if we need his shared 1.17 changes

PR #88, opened 2023-01-27
Updated the deploy docs

thomasm-ttd:
Update the config for aws integration, azure and gcp

### DIFF
--- a/scripts/gcp/README.md
+++ b/scripts/gcp/README.md
@@ -78,6 +78,9 @@ As shown in the example below, it first disables remote SSH access, and then tel
 COS VM to docker pull certified UID2 operator docker image from UID2 project's official
 Container Registry and run the UID2 operator container as a systemd service.
 
+The UID2_ENCLAVE_API_TOKEN, UID2_ENCLAVE_IMAGE_ID and GHCR_RO_ACCESS_TOKEN will all be set in the file that 
+you are provided with. There is no need to edit them manually.
+
 ```
 #cloud-config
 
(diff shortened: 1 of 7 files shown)


scripts/gcp/README.md
```diff
@@ -133,139 +123,66 @@ write_files:
...
-as custom metadata under the key `user-data` when creating the VM instance. This `user-data`
-metadata will be read and interpreted by the Container-Optimized OS (COS) VM disk during
-booting.
-
-As shown in the template below, it first disables remote SSH access, and then tells
-COS VM to docker pull certified UID2 operator docker image from UID2 project's official
-Container Registry and run the UID2 operator container as a systemd service.
-
-```
-#cloud-config
```
scong-ttd, line 200: Do we need to keep the cloud-init here?

scripts/gcp/README.md
```diff
@@ -112,11 +101,12 @@ write_files:
     [Service]
     Environment="UID2_ENCLAVE_API_TOKEN=<INPUT_API_TOKEN>"
     Environment="UID2_ENCLAVE_IMAGE_ID=<INPUT_IMAGE_ID>"
+    Environment="GHCR_RO_ACCESS_TOKEN=<GHCR_TOKEN>"
```
scong-ttd, line 104: any explanation what this token is and where for clients to get it?

thomasm-ttd: The clients don't get it, we will provide it. I will add a note about that

scripts/gcp/README.md
thomasm-ttd: I thought having an example and discussing it was useful, but happy to remove the example if you don't think it is useful

scong-ttd: Oh I thought this was a deletion, but actually confusion from `git diff`.

scong-ttd (approved):

PR #90, opened 2023-01-27
Set optout_s3_path_compat to false everywhere

scong-ttd:
optout_s3_path_compat was a backward-compatibility hack introduced in optout v1 -> v2 transition period. It should be enabled nowhere.

Also removed scripts/aws/config.json as it is not used.

Keeping the current version in pom.xml as Operator code stays the same.


scong-ttd:
We should consider clean up the config files.

PR #97, opened 2023-02-07
Replace `io.vertx.core.logging` with `org.slf4j`

cYKatherine:


cYKatherine:
I know some of you might be wondering that:
- in `Dockerfile` we still have `-Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory`
- in `entrypoint.sh` we still have `SETUP_LOKI_LINE="-Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory -Dlogback.configurationFile=./conf/logback.loki.xml"`

Those exist because we wanted Vert.x to use SLF4J instead of JUL for it’s own logging. The `vertx.logger-delegate-factory-class-name` controls where calls to `io.vertx.core.logging.Logger.*` end up. ([reference](https://vertx.io/docs/3.9.14/vertx-core/java/#_using_another_logging_framework)).

We will keep them as it is, (even in @gmsdelmundo-ttd 's change of upgrading vertx). The reason being:
- only `Logger` and `LoggerFactory` will be deprecated - `SLF4JLogDelegateFactory` is safe. ([reference](https://vertx.io/docs/apidocs/io/vertx/core/logging/package-summary.html))
- Even we change our application code to `org.slf4j.logging`, the Vert.x framework still uses the `io.vertx.core.logging.Logger` API, and we need to make sure them using the `slf4j` logging

cc: @mcollins-ttd

mcollins-ttd (approved):

scong-ttd:
Food for thought: is it valuable to create an abstraction on our own, so we can move to another logging library in the case of vulnerability? (This change is small enough so probably fine)

thomasm-ttd (approved):

PR #103, opened 2023-02-15
Upgraded Vert.x to 4.3.8

ghost:


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -719,7 +716,7 @@ private void handleIdentityMapV1(RoutingContext rc) {
             final MappedIdentity mappedIdentity = this.idService.map(input.toUserIdentity(this.identityScope, 0, now), now);
             final JsonObject jsonObject = new JsonObject();
             jsonObject.put("identifier", input.getProvided());
-            jsonObject.put("advertising_id", mappedIdentity.advertisingId);
+            jsonObject.put("advertising_id", EncodingUtils.toBase64String(mappedIdentity.advertisingId));
```
thomasm-ttd, line 719: Does this change what is actually written? Was it base64encoded previously?
Same for below where a similar change is made

cYKatherine: Is it related to https://github.com/IABTechLab/uid2-shared/pull/56/files#r1120950496?

gmsdelmundo-ttd: @cYKatherine nope not related

@thomasm-ttd Actually, for this one, I think we may not actually need it since `advertisingId` is handled differently than the encryption key in uid2-admin, I'll revert this particular change

cYKatherine (approved):

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
gmsdelmundo-ttd: Nvm, we actually need all the Base64 encodes for `JsonObject.put()` with `byte[]` data types because of the Vert.x 4 deprecation of their `JsonObject` implementation

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -978,7 +975,7 @@ private void handleIdentityMapV2(RoutingContext rc) {
                     final MappedIdentity mappedIdentity = idService.map(input.toUserIdentity(this.identityScope, 0, now), now);
                     final JsonObject resp = new JsonObject();
                     resp.put("identifier", input.getProvided());
-                    resp.put("advertising_id", mappedIdentity.advertisingId);
+                    resp.put("advertising_id", EncodingUtils.toBase64String(mappedIdentity.advertisingId));
```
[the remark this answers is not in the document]
thomasm-ttd: Based on discussions with @gmsdelmundo , this toBase64String was included in the .put method in Vertx 3, but not in 4, so we need to do it now

thomasm-ttd (approved):

PR #104, opened 2023-02-16
Cbc UI d2 705 sharing endpoint for id readers

cody-constine-ttd:


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -274,13 +277,19 @@ public void handleKeysSharing(RoutingContext rc) {
 
             final List<EncryptionKey> keyStore = getEncryptionKeys();
 
+            MissingAclMode mode = MissingAclMode.DENY_ALL;
+            IRoleAuthorizable<Role> roleAuthorize = (IRoleAuthorizable<Role>) rc.data().get(API_CLIENT_PROP);
```
jon8787, line 281: should we put a comment here, or where data is added, to warn about how this could break if a different type is added?

PR #105, opened 2023-03-01
New identityMap API with micro benchmark

scong-ttd:
Changes:
1. deprecate UIDOperatorService.map method
2. create UIDOperatorService.identityMap method and MapRequest (similar to IdentityRequest)
3. add microbenchmark to estimate the performance impact

### DIFF
--- a/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
+++ b/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
@@ -70,11 +70,14 @@ public class UIDOperatorVerticle extends AbstractVerticle{
     private IUIDOperatorService idService;
     private final Map<String, DistributionSummary> _identityMapMetricSummaries = new HashMap<>();
     private final Map<String, DistributionSummary> _refreshDurationMetricSummaries = new HashMap<>();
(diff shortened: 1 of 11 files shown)


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -938,7 +956,14 @@ private void handleIdentityMapBatchV1(RoutingContext rc) {
             for (int i = 0; i < count; ++i) {
                 final InputUtil.InputVal input = inputList[i];
                 if (input != null && input.isValid()) {
-                    final MappedIdentity mappedIdentity = this.idService.map(input.toUserIdentity(this.identityScope, 0, now), now);
+                    final MappedIdentity mappedIdentity = this.idService.mapIdentity(
+                            new MapRequest(
+                                    input.toUserIdentity(this.identityScope, 0, now),
+                                    IdentityMapPolicy.defaultPolicy(),
+                                    now));
+                    if (mappedIdentity.isOptedOut())
```
cody-constine-ttd, line 964: Should we put something in the response here to indicate the the requested user was opted out?

scong-ttd: added a separate list "failed" next to "mapped"

PR #107, opened 2023-03-06
Removed query string from URL for metrics

gmsdelmundo:


src/main/java/com/uid2/operator/Main.java
```diff
@@ -353,21 +353,21 @@ private static void setupMetrics(MicrometerMetricsOptions metricOptions) {
...
-                .meterFilter(MeterFilter.deny(id ->
-                    id.getName().startsWith(MetricsDomain.HTTP_SERVER.getPrefix()) &&
-                    Objects.equals(id.getTag(Label.HTTP_CODE.toString()), "404")))
-                // adding common labels
-                .commonTags("application", "uid2-operator");
+                    // providing common renaming for prometheus metric, e.g. "hello.world" to "hello_world"
+                    .meterFilter(new PrometheusRenameFilter())
+                    .meterFilter(MeterFilter.replaceTagValues(Label.HTTP_PATH.toString(), actualPath -> {
+                        try {
+                            return HttpUtils.normalizePath(actualPath).split("\\?")[0];
```
scong-ttd, line 360: why "\\?" instead of "?"

scong-ttd (commented):
NIT: since we don't force a code formatting let's follow the existing formatting and avoid whitespace changes :)

src/main/java/com/uid2/operator/Main.java
gmsdelmundo: `.split()` takes in regex, so ? needs to be escaped

gmsdelmundo:
> NIT: since we don't force a code formatting let's follow the existing formatting and avoid whitespace changes :)

Updated

src/main/java/com/uid2/operator/Main.java
scong-ttd: 👍

scong-ttd (approved):

PR #132, opened 2023-05-08
Jon UI d2 1010 operator logs not showing participant in v2 ap is

jon8787:


src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -1210,7 +1225,7 @@ void identityMapBothEmailAndHashSpecified(Vertx vertx, VertxTestContext testCont
         setupKeys();
         get(vertx, "v1/identity/map?email=" + emailAddress + "&email_hash=" + urlEncode(emailHash), ar -> {
             assertTrue(ar.succeeded());
-            HttpResponse response = ar.result();
+            HttpResponse<Buffer> response = ar.result();
```
sunnywu, line 1228: why is <Buffer> needed here and below?

jon8787: fixing an intellij code analysis warning to not use raw types ( https://stackoverflow.com/questions/2770321/what-is-a-raw-type-and-why-shouldnt-we-use-it )

mcollins-ttd (approved):

PR #134, opened 2023-05-08
Measure 90th, 95th, and 99th percentile latencies

mcollins-ttd:


pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-operator</artifactId>
-    <version>2.1.5</version>
```
sunnywu, line 9: https://semver.org/
This doesn't count as bufix so i would suggest to increment the minor version i.e. from 2.1.5 to 2.2.0 instead of to 2.1.6.
But can do it next time not too fuss for this MR

sunnywu (approved):

PR #135, opened 2023-05-26
advertising token v4 support

jon8787:


src/main/java/com/uid2/operator/service/Uid2Base64UrlCoder.java
```diff
@@ -0,0 +1,41 @@
+package com.uid2.operator.service;
+
+import java.util.Base64;
+
```
sunnywu, line 4: maybe add a comment/link to the code/file in UID2 SDK For java about they are using same encode/decoder?

src/test/java/com/uid2/operator/TokenEncodingTest.java
```diff
@@ -32,62 +35,75 @@ public TokenEncodingTest() throws Exception {
...
-        Assert.assertEquals(TokenVersion.V2, decoded.version);
-        Assert.assertEquals(token.createdAt, decoded.createdAt);
-        Assert.assertEquals(token.expiresAt.plusSeconds(60), decoded.expiresAt);
-        Assert.assertTrue(token.userIdentity.matches(decoded.userIdentity));
-        Assert.assertEquals(token.userIdentity.privacyBits, decoded.userIdentity.privacyBits);
-        Assert.assertEquals(token.userIdentity.establishedAt, decoded.userIdentity.establishedAt);
-        Assert.assertEquals(token.publisherIdentity.siteId, decoded.publisherIdentity.siteId);
+        assertEquals(tokenVersion, decoded.version);
+        assertEquals(token.createdAt, decoded.createdAt);
+        int addSeconds = (tokenVersion == TokenVersion.V2) ? 60 : 0; //todo: why is there a 60 second buffer in encodeV2() but not in encodeV3()?
```
sunnywu, line 63: are we keeping this as is or make it consistent?

sunnywu:
Could you update this page https://atlassian.thetradedesk.com/confluence/display/UID2/UID2+Version+Numbers+of+Various+Things on refresh token version (which is at v3 not v4)?

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -68,7 +68,11 @@ public UIDOperatorService(JsonObject config, IOptOutStore optOutStore, ISaltProv
             throw new IllegalStateException(REFRESH_TOKEN_EXPIRES_AFTER_SECONDS + " must be >= " + REFRESH_IDENTITY_TOKEN_AFTER_SECONDS);
         }
 
-        this.advertisingTokenVersion = config.getBoolean("advertising_token_v3", false) ? TokenVersion.V3 : TokenVersion.V2;
+        if (config.getBoolean("advertising_token_v4", false)) {
+            this.advertisingTokenVersion = TokenVersion.V4;
+        } else {
+            this.advertisingTokenVersion = config.getBoolean("advertising_token_v3", false) ? TokenVersion.V3 : TokenVersion.V2;
```
sunnywu, line 74: should we throw an exception if both advertising_token_v3/advertising_token_v4 are set to true?

src/test/java/com/uid2/operator/TokenEncodingTest.java
```diff
@@ -32,62 +35,75 @@ public TokenEncodingTest() throws Exception {
...
+
+        final byte[] firstLevelHash = TokenUtils.getFirstLevelHashFromIdentity("test@example.com", "some-salt");
+
+        final RefreshToken token = new RefreshToken(tokenVersion,
             now,
             now.plusSeconds(360),
             new OperatorIdentity(101, OperatorType.Service, 102, 103),
             new PublisherIdentity(111, 112, 113),
-            new UserIdentity(IdentityScope.UID2, IdentityType.Email, "some-id".getBytes(StandardCharsets.UTF_8), 121, now, now.minusSeconds(122))
+            new UserIdentity(IdentityScope.UID2, IdentityType.Email, firstLevelHash, 121, now, now.minusSeconds(122))
```
sunnywu, line 51: the new TokenUtils.getFirstLevelHashFromIdentity method used to generate firstLevelHash doesn't seem to (indirectly) define StandardCharsets.UTF_8, is this ok?

conf/local-config.json
```diff
@@ -9,10 +9,11 @@
   "refresh_token_expires_after_seconds": 86400,
   "refresh_identity_token_after_seconds": 900,
   "advertising_token_v3": false,
+  "advertising_token_v4": true,
   "refresh_token_v3": false,
   "identity_v3": false,
   "identity_scope": "uid2",
-  "enable_v2_encryption": false,
+  "enable_v2_encryption": true,
```
mcollins-ttd, line 16: Is this intentional? People might have to update their dev workflow around this change.

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
```diff
@@ -144,17 +145,38 @@ private RefreshToken decodeRefreshTokenV3(Buffer b, byte[] bytes) {
...
+        byte[] headerBytes = isBase64UrlEncoding ? Uid2Base64UrlCoder.decode(headerStr) : Base64.getDecoder().decode(headerStr);
+
+        if (headerBytes[0] == TokenVersion.V2.rawVersion) {
+            final byte[] bytes = EncodingUtils.fromBase64(base64AdvertisingToken);
+            final Buffer b = Buffer.buffer(bytes);
             return decodeAdvertisingTokenV2(b);
+        }
+
+        //Java's byte is signed, so we convert to unsigned before checking the enum
+        int unsignedByte = ((int) headerBytes[1]) & 0xff;
```
mcollins-ttd, line 164:
```suggestion
        int unsignedByte = Byte.toUnsignedInt(headerBytes[1]);
```

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
```diff
@@ -144,17 +145,38 @@ private RefreshToken decodeRefreshTokenV3(Buffer b, byte[] bytes) {
...
+            bytes = EncodingUtils.fromBase64(base64AdvertisingToken);
+            tokenVersion = TokenVersion.V3;
+        } else if (unsignedByte == TokenVersion.V4.rawVersion) {
+            bytes = Uid2Base64UrlCoder.decode(base64AdvertisingToken);  //same as V3 but use Base64URL encoding
+            tokenVersion = TokenVersion.V4;
         } else {
             throw new IllegalArgumentException("Invalid advertising token version");
         }
+
+        final Buffer b = Buffer.buffer(bytes);
```
mcollins-ttd, line 178: Nit: `Buffer.buffer(byte[])` copies the bytes. Since we own the `byte[]`, we can avoid the copying with:

```suggestion
        final Buffer b = Buffer.buffer(Unpooled.wrappedBuffer(bytes));
```

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
```diff
@@ -225,17 +247,21 @@ public AdvertisingToken decodeAdvertisingTokenV3(Buffer b, byte[] bytes) {
...
         );
     }
 
     public byte[] encode(RefreshToken t, Instant asOf) {
         final EncryptionKey serviceKey = this.keyStore.getSnapshot().getRefreshKey(asOf);
 
-        return t.version == TokenVersion.V3
-                ? encodeV3(t, serviceKey)
-                : encodeV2(t, serviceKey);
+        if (t.version == TokenVersion.V2) {
```
mcollins-ttd, line 258: Nit: I'd prefer a `switch`.

src/main/java/com/uid2/operator/service/UIDOperatorService.java
jon8787: I think I prefer "newest token version wins", as that's still fairly intuitive behavior

conf/local-config.json
jon8787: will revert. But it's interesting that here we want "not production-like" but elsewhere we want "production-like" https://github.com/IABTechLab/uid2-operator/pull/135#discussion_r1208839036

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
jon8787: note that this code is a copy/paste and will be duplicated 3 times :( 
@sunnywu thoughts?

conf/local-config.json
sunnywu: any reason why we don't want to keep it prod like?

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
sunnywu: is this copy/paste in multiple repos/projects? if so maybe do this as ContinuousImprovements in another ticket for all repos.
(i assume this is not a bugfix but syntactic improvement)

src/test/java/com/uid2/operator/TokenEncodingTest.java
jon8787: yep as discussed, getFirstLevelHashFromIdentity is used in prod to generate raw UIDs

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
jon8787: Yes, this code is copied here to Operator from Java SDK, and will be copied further to Validator V2.
Shall we leave this as part of UID2-1206 ?

sunnywu: yeh i agree on add this improvement requirement into UID2-1206

jon8787: @mcollins-ttd fyi this code came from https://github.com/IABTechLab/uid2-client-java/blob/0220ef43c1661ecf3b8f4ed2db524e2db31c06b5/src/main/java/com/uid2/client/Uid2Encryption.java#L37 , I'll add that as a comment

src/test/java/com/uid2/operator/TokenEncodingTest.java
jon8787: pending discussion - can leave it for another ticket if necessary as I haven't changed behavior

conf/local-config.json
jon8787: @mcollins-ttd ?

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
jon8787: as above, we can leave this as part of UID2-1206
(noting this is for /token/validate which is rarely called)

conf/local-config.json
mcollins-ttd: I'd like to keep the value as-is for now, as it's not related to the V4 token support. Can discuss with the team if it will break anyone's workflow.

PR #148, opened 2023-07-11
Add KeySets Support to Operator

zaiweidu:


scripts/aws/README.md
```diff
@@ -48,6 +48,8 @@ UID2 Operator application reads configuration from [AWS Secrets Manager](https:/
   "keys_metadata_path": "https://core-integ.uidapi.com/key/refresh",
   "salts_metadata_path": "https://core-integ.uidapi.com/salt/refresh",
   "keys_acl_metadata_path": "https://core-integ.uidapi.com/key/acl/refresh",
+  "keyset_metadata_path": "https://core-integ.uidapi.com/keyset/refresh",
+  "keyset_keys_metadata_path": "https://core-integ.uidapi.com/keyset_keys/refresh",
```
cody-constine-ttd, line 52: the path is keyset-keys

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -290,13 +290,13 @@ public void handleKeysSharing(RoutingContext rc) {
...
+            for (KeysetKey key: keysetKeyStore) {
                 JsonObject keySet = new JsonObject();
-                if(clientKey.getSiteId() == key.getSiteId()) {
+                if(clientKey.getSiteId() == this.keysetProvider.getKeyset(key.getKeysetId()).getSiteId()) {
                     keySet.put("keyset_id", DEFAULT_KEYSET_ID);
-                } else if (key.getSiteId() == -1) {
+                } else if (this.keysetProvider.getKeyset(key.getKeysetId()).getSiteId() == -1) {
                     keySet.put("keyset_id", DEFAULT_MASTER_KEYSET_ID);
-                } else if (!acls.canClientAccessKey(clientKey, key, mode)) {
+                } else if (!keysets.canClientAccessKey(clientKey, key)) {
```
cody-constine-ttd, line 299: We might still need the mode here, because ID_READERs can get null keysets

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -290,13 +290,13 @@ public void handleKeysSharing(RoutingContext rc) {
                 mode = MissingAclMode.ALLOW_ALL;
             }
 
-            for (EncryptionKey key: keyStore) {
+            for (KeysetKey key: keysetKeyStore) {
                 JsonObject keySet = new JsonObject();
-                if(clientKey.getSiteId() == key.getSiteId()) {
+                if(clientKey.getSiteId() == this.keysetProvider.getKeyset(key.getKeysetId()).getSiteId()) {
                     keySet.put("keyset_id", DEFAULT_KEYSET_ID);
```
cody-constine-ttd, line 296: This should be the clients default Keyset ID now

jon8787:
There are a few EncryptionKey references in EncryptionTest. Can we replace these with KeysetKey?

conf/docker-config.json
```diff
@@ -27,6 +27,8 @@
   "clients_metadata_path": "/com.uid2.core/test/clients/metadata.json",
   "keys_metadata_path": "/com.uid2.core/test/keys/metadata.json",
   "keys_acl_metadata_path": "/com.uid2.core/test/keys_acl/metadata.json",
+  "keyset_metadata_path": "/com.uid2.core/test/keyset/metadata.json",
```
jon8787, line 30: why do we not remove "keys_metadata_path" and "keys_acl_metadata_path" from all these json files?

scripts/aws/make_config.py
```diff
@@ -39,6 +39,8 @@ def apply_override(config, overrides, key, type):
   apply_override(config, integ_config, 'clients_metadata_path', str)
   apply_override(config, integ_config, 'keys_metadata_path', str)
   apply_override(config, integ_config, 'keys_acl_metadata_path', str)
+  apply_override(config, integ_config, 'keyset_metadata_path', str)
```
jon8787, line 42: should we remove the 2 "keys" lines above?

src/main/java/com/uid2/operator/monitoring/OperatorMetrics.java
```diff
@@ -40,15 +44,15 @@ public void setup() {
...
+        keysetProvider.getSnapshot().getAllKeysets().values().stream()
                 .map(k -> k.getSiteId()).distinct()
                 .filter(s -> !encryptionKeyGaugesBySiteId.contains(s))
                 .forEachOrdered(siteId -> {
                     encryptionKeyGaugesBySiteId.add(siteId);
                     Gauge
                             .builder("uid2_encryption_key_activates", () -> {
                                 final Instant now = Instant.now();
-                                final EncryptionKey key = keyStore.getSnapshot().getActiveSiteKey(siteId, now);
+                                final KeysetKey key = EncryptionKeyUtil.getDefaultKeysetActiveKey(keysetKeyStore.getSnapshot(), keysetProvider, siteId, now);
```
jon8787, line 55: do we also need to deal with non-default active keys?

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
```diff
@@ -15,23 +16,27 @@
...
+        this.keysetKeyStore = keysetKeyStore;
+        this.keysetProvider = keysetProvider;
     }
 
     public byte[] encode(AdvertisingToken t, Instant asOf) {
-        final EncryptionKey masterKey = this.keyStore.getSnapshot().getMasterKey(asOf);
-        final EncryptionKey siteEncryptionKey = EncryptionKeyUtil.getActiveSiteKey(
-                this.keyStore.getSnapshot(), t.publisherIdentity.siteId, Const.Data.AdvertisingTokenSiteId, asOf);
+        final KeysetKey masterKey = EncryptionKeyUtil.getActiveKeysetKey(
+                this.keysetKeyStore.getSnapshot(), this.keysetProvider, Const.Data.MasterKeySiteId, Const.Data.MasterKeySiteId, asOf);
```
jon8787, line 29: I think we should have a different function where there is no fallback site id, because that's needed both here and for the refresh key

src/main/java/com/uid2/operator/service/EncryptionKeyUtil.java
```diff
@@ -1,17 +1,37 @@
...
-            throw new RuntimeException(String.format("cannot get active site key with ID %d or %d", siteId, fallbackSiteId));
+            throw new RuntimeException(String.format("cannot get active key in default keyset with ID %d or %d", siteId, fallbackSiteId));
         }
         return key;
     }
+
+    public static KeysetKey getDefaultKeysetActiveKey(IKeysetKeyStore.IkeysetKeyStoreSnapshot keysetKeyStoreSnapshot, RotatingKeysetProvider keysetProvider, int siteId, Instant now) {
+        List<Keyset> keysets = keysetProvider.getSnapshot().getAllKeysets().values().stream()
+                .filter(s -> s.getSiteId() == siteId)
+                .filter(d -> d.isDefault() == true)
```
jon8787, line 25: can remove "== true"?

src/main/java/com/uid2/operator/service/EncryptionKeyUtil.java
```diff
@@ -1,17 +1,37 @@
...
+                .filter(e -> e.isEnabled() == true)
+                .collect(Collectors.toList());
+
+        Keyset defaultKeyset = !keysets.isEmpty() ? keysets.get(0) : null;
```
jon8787, line 29: what happens if keysets > 1 ?

src/main/java/com/uid2/operator/service/EncryptionKeyUtil.java
```diff
@@ -1,17 +1,37 @@
...
+        if (defaultKeyset == null) {
+            return null;
```
jon8787, line 31: should we log an error here?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -290,13 +284,18 @@ public void handleKeysSharing(RoutingContext rc) {
                 mode = MissingAclMode.ALLOW_ALL;
             }
 
-            for (EncryptionKey key: keyStore) {
+            int defaultMasterKeysetId = DEFAULT_MASTER_KEYSET_ID;
+            int defaultKeysetId = DEFAULT_KEYSET_ID;
```
jon8787, line 288: I think the above lines are not correct. We shouldn't be using these consts anymore, and should delete them from the repo

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -290,13 +284,18 @@ public void handleKeysSharing(RoutingContext rc) {
...
-                    keySet.put("keyset_id", DEFAULT_KEYSET_ID);
-                } else if (key.getSiteId() == -1) {
-                    keySet.put("keyset_id", DEFAULT_MASTER_KEYSET_ID);
-                } else if (!acls.canClientAccessKey(clientKey, key, mode)) {
+                Keyset keyset = this.keysetProvider.getKeyset(key.getKeysetId());
+                if(clientKey.getSiteId() == keyset.getSiteId() && keyset.isEnabled()) {
+                    if (keyset.isDefault()) defaultKeysetId = key.getKeysetId();
+                    keySet.put("keyset_id", key.getKeysetId());
+                } else if (keyset.getSiteId() == Const.Data.MasterKeySiteId && keyset.isEnabled()) {
+                    if (keyset.isDefault()) defaultMasterKeysetId = key.getKeysetId();
```
jon8787, line 296: Do we have non-default master keysets?

src/main/resources/com.uid2.core/test/keyset/keyset.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
```
jon8787, line 2: can we remove keys.json and keys_acl.json?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -105,8 +106,10 @@ public class UIDOperatorVerticleTest {
     @BeforeEach
     void deployVerticle(Vertx vertx, VertxTestContext testContext) {
         mocks = MockitoAnnotations.openMocks(this);
-        when(keyStore.getSnapshot()).thenReturn(keyStoreSnapshot);
-        when(keyAclProvider.getSnapshot()).thenReturn(keyAclProviderSnapshot);
+        when(keysetKeyStore.getSnapshot()).thenReturn(keysetKeyStoreSnapshot);
+        when(keysetProvider.getSnapshot()).thenReturn(keysetProviderSnapshot);
+
+        //when(keysetProvider.getSnapshot()).thenReturn(keysetProviderSnapshot);
```
jon8787, line 112: could remove this

scripts/aws/make_config.py
zaiweidu: test

scripts/aws/README.md
zaiweidu: Fixed the path.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -290,13 +290,13 @@ public void handleKeysSharing(RoutingContext rc) {
...
+                if(clientKey.getSiteId() == this.keysetProvider.getKeyset(key.getKeysetId()).getSiteId()) {
                     keySet.put("keyset_id", DEFAULT_KEYSET_ID);
-                } else if (key.getSiteId() == -1) {
+                } else if (this.keysetProvider.getKeyset(key.getKeysetId()).getSiteId() == -1) {
                     keySet.put("keyset_id", DEFAULT_MASTER_KEYSET_ID);
-                } else if (!acls.canClientAccessKey(clientKey, key, mode)) {
+                } else if (!keysets.canClientAccessKey(clientKey, key)) {
                     continue;
                 }
                 keySet.put("id", key.getId());
```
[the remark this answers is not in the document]
zaiweidu: Set the actual KeysetId.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
zaiweidu: Replaced with keysetId of default keyset.

conf/docker-config.json
zaiweidu: Removed used "keys_metadata_path" and "keys_acl_metadata_path" keys.

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
zaiweidu: Added a wrapper function that only takes 1 siteId (no fallback).

src/main/java/com/uid2/operator/service/EncryptionKeyUtil.java
zaiweidu: Updated in the latest commit.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
zaiweidu: Replaced with masterKeysetId and defaultKeysetId.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
zaiweidu: Removed the .isDefault() check.

src/main/resources/com.uid2.core/test/keyset/keyset.json
zaiweidu: Removed keys.json and keys_acl.json.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
zaiweidu: Removed.

src/main/java/com/uid2/operator/service/EncryptionKeyUtil.java
zaiweidu: One site can only have 1 default Keyset.

src/main/java/com/uid2/operator/service/EncryptionKeyUtil.java
```diff
@@ -1,17 +1,45 @@
...
+            .filter(s -> s.getSiteId() == siteId && s.isDefault() && s.isEnabled())
+            .collect(Collectors.toList());
+
+        if (keysets.size() > 1) {
+            throw new RuntimeException(String.format("Multiple default keysets are enabled with SITE ID %d.", siteId));
+        }
+
+        Keyset defaultKeyset = !keysets.isEmpty() ? keysets.get(0) : null;
+        if (defaultKeyset == null) {
+            return null;
```
jon8787, line 39: If people call this method directly and there's no key, it will return null, but if they call the fallback method it will throw. Is this desired?

src/main/java/com/uid2/operator/service/EncryptionKeyUtil.java
jon8787: That's an assumption. If the assumption is found to be false, should we log an error?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -271,44 +266,53 @@ public void handleKeysRequest(RoutingContext rc) {
...
-                } else if (key.getSiteId() == -1) {
-                    keySet.put("keyset_id", DEFAULT_MASTER_KEYSET_ID);
-                } else if (!acls.canClientAccessKey(clientKey, key, mode)) {
+            KeysetKey masterKey = EncryptionKeyUtil.getActiveKeyBySiteId(this.keysetKeyStore.getSnapshot(), keysetSnapshot, Const.Data.MasterKeySiteId, Instant.now());
+            if (masterKey == null) {
+                throw new RuntimeException(String.format("Cannot get active master key with SITE ID %d.", Const.Data.MasterKeySiteId));
+            }
+
+            // defaultKeysetId allows calling sdk.Encrypt(rawUid) without specifying the keysetId
+            int defaultKeysetId = EncryptionKeyUtil.getActiveKeyBySiteIdWithFallback(this.keysetKeyStore.getSnapshot(), keysetSnapshot, clientKey.getSiteId(), Const.Data.AdvertisingTokenSiteId, Instant.now()).getKeysetId();
```
jon8787, line 286: for sharing, we'd never want to fallback, so this should be getActiveKeyBySiteId()

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -271,44 +266,53 @@ public void handleKeysRequest(RoutingContext rc) {
...
+
+            // include 'keyset_id' field, if:
+            //   (a) a key belongs to caller's site
+            //   (b) a key belongs to master_keyset
+            // otherwise, when a key is accessible by caller, the key can be used for decryption only. skip 'keyset_id' field.
+            for (KeysetKey key: keysetKeyStore) {
+                JsonObject keyObj = new JsonObject();
+                Keyset keyset = this.keysetProvider.getSnapshot().getKeyset(key.getKeysetId());
```
jon8787, line 294: 1. We've already retrieved keysetSnapshot, why are we getting it again here?
2. Probably not correct to call getSnapshot() here because we could get a different snapshot in any iteration of the loop. So when we move to KeyManager, will need to solve that (either expose getSnapshot() after all, or extract this data from KeyManager with a new method)

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
```diff
@@ -15,23 +16,30 @@
...
     }
 
     public byte[] encode(AdvertisingToken t, Instant asOf) {
-        final EncryptionKey masterKey = this.keyStore.getSnapshot().getMasterKey(asOf);
-        final EncryptionKey siteEncryptionKey = EncryptionKeyUtil.getActiveSiteKey(
-                this.keyStore.getSnapshot(), t.publisherIdentity.siteId, Const.Data.AdvertisingTokenSiteId, asOf);
+        final KeysetKey masterKey = EncryptionKeyUtil.getActiveKeyBySiteId(
+                this.keysetKeyStore.getSnapshot(), keysetProvider.getSnapshot(), Const.Data.MasterKeySiteId, asOf);
+        if (masterKey == null) {
+            throw new RuntimeException(String.format("Cannot get active master key with SITE ID %d.", Const.Data.MasterKeySiteId));
```
cody-constine-ttd, line 31: We should probably log an error here

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
cody-constine-ttd: Can't we set defaultKeysetId as we look through the keys anyways? We are doing an extra loop through all the keys here

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1416,11 +1423,18 @@ private JsonObject toJsonV1(IdentityTokens t) {
         return json;
     }
 
-    private JsonArray toJson(List<EncryptionKey> keys, ClientKey clientKey, IKeysAclSnapshot acls) {
+    private JsonArray toJson(List<KeysetKey> keys, ClientKey clientKey, RotatingKeysetProvider keysetProvider) {
+        /* Reviewers to confirm: whether we need to check MissingAclMode by client's role.
```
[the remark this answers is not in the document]
cody-constine-ttd: I don't think a toJson method should have logic about who can access keys TBH, but if we want to keep this the way it is, then it probably needs the mode

src/test/java/com/uid2/operator/RotatingKeysetKeyStoreTest.java
```diff
@@ -0,0 +1,21 @@
+package com.uid2.operator;
```
cody-constine-ttd, line 1: Why are we testing RotatingKeysetKeyStore here? There is already test for it in Shared

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -148,8 +155,38 @@ private static byte[] makeAesKey(String prefix) {
...
+    private void addKeysets(Keyset... keysets) {
+        // keysetMap: keysetId -> keyset
+        Map<Integer, Keyset> keysetMap = Arrays.stream(keysets).collect(Collectors.toMap(s -> s.getKeysetId(), s -> s));
+        when(keysetProviderSnapshot.getKeyset(anyInt())).then((i) -> {
+            return keysetMap.get(i.getArgument(0, Integer.class));
+        });
+        when(keysetProviderSnapshot.getAllKeysets()).thenReturn(keysetMap);
+    }
+
+    private void linkKeysToKeysets(KeysetKey[] keys, Keyset[] keysets) {
```
cody-constine-ttd, line 176: I know you explained this function to me, but I still don't understand its purpose. Keys and Keysets should be linked by keyset ID

cody-constine-ttd (commented):
Can't currently run because the shared version is uncorrect in the pom.xml

conf/default-config.json
```diff
@@ -21,11 +21,15 @@
...
+  "keysets_metadata_path": "keysets/metadata.json",
+  "keyset_keys_metadata_path": "keyset_keys/metadata.json",
   "salts_metadata_path": "salts/metadata.json",
   "optout_metadata_path": null,
   "optout_inmem_cache": false,
   "enclave_platform": null,
   "failure_shutdown_wait_hours": 120,
-  "sharing_token_expiry_seconds": 2592000
+  "sharing_token_expiry_seconds": 2592000,
+  "master_key_keyset_id": 1,
```
jon8787, line 32: Why do we put these ids in config rather than a const? To me, the main reason to do this would be if we wanted different values between environments. But I think that's something we don't want here.

Site ids don't work like this, eg Data.MasterKeySiteId does not come from a config

src/main/java/com/uid2/operator/model/KeyManager.java
```diff
@@ -0,0 +1,70 @@
...
+        return key;
+    }
+
+    // Retrieve an active key from default keyset by caller's site id.
+    public KeysetKey getActiveKeyBySiteId(int siteId, Instant now) {
+        List<Keyset> keysets = this.keysetProvider.getSnapshot().getAllKeysets().values().stream()
+                .filter(s -> s.isEnabled() && s.isDefault() && s.getSiteId() == siteId)
+                .collect(Collectors.toList());
+        if (keysets.isEmpty()) {
+            return null;
```
jon8787, line 39: Concerns me that we don't log or throw here. I think there's a bug in one of the calling methods, because it doesn't throw or check for null.
Can we add getMasterKey and getRefreshKey methods in KeyManager which throw on failure?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -416,23 +454,41 @@ protected void setupSalts() {
...
-        EncryptionKey siteKey = new EncryptionKey(keyId, makeAesKey("siteKey" + siteId), Instant.now().minusSeconds(7), Instant.now(), Instant.now().plusSeconds(10), siteId);
-        when(keyStoreSnapshot.getActiveSiteKey(eq(siteId), any())).thenReturn(siteKey);
-        when(keyStoreSnapshot.getKey(keyId)).thenReturn(siteKey);
+        KeysetKey masterKey = new KeysetKey(101, makeAesKey("masterKey"), Instant.now().minusSeconds(7), Instant.now(), Instant.now().plusSeconds(10), 1);
+        KeysetKey siteKey = new KeysetKey(102, makeAesKey("siteKey"), Instant.now().minusSeconds(7), Instant.now(), Instant.now().plusSeconds(10), 2);
+        KeysetKey refreshKey = new KeysetKey(103, makeAesKey("refreshKey"), Instant.now().minusSeconds(7), Instant.now(), Instant.now().plusSeconds(10), 3);
+        when(keysetProviderSnapshot.canClientAccessKey(any(), any(), any())).thenReturn(true);
+        when(keysetKeyStoreSnapshot.getActiveKey(eq(1), any())).thenReturn(masterKey);
+        when(keysetKeyStoreSnapshot.getActiveKey(eq(2), any())).thenReturn(refreshKey);
+        when(keysetKeyStoreSnapshot.getActiveKey(eq(3), any())).thenReturn(siteKey);
```
jon8787, line 463: the above is another reason we should use consts for refresh keyset id etc - the above looks like it depends on a specific config being used

src/main/java/com/uid2/operator/model/KeyManager.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+        if (keysets.size() > 1) {
+            throw new IllegalArgumentException("Multiple default keysets are enabled with SITE ID " + siteId);
+        }
+        return getActiveKey(keysets.get(0).getKeysetId(), now);
+    }
+
+    public KeysetKey getActiveKey(int keysetId, Instant now) {
```
jon8787, line 47: can this be private?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -317,11 +322,13 @@ public void handleKeysSharing(RoutingContext rc) {
...
     }
 
-    private List<EncryptionKey> getEncryptionKeys() {
-        final List<EncryptionKey> keyStore = this.keyStore.getSnapshot().getActiveKeySet()
-                .stream().filter(k -> k.getSiteId() != Const.Data.RefreshKeySiteId)
+    private List<KeysetKey> getKeysetKeys() {
+        Map<Integer, Keyset> keysetMap = this.keyManager.getAllKeysets();
+        List<KeysetKey> keys = this.keyManager.getActiveKeysetKeys();
+        final List<KeysetKey> keysetKeyStore = keys
+                .stream().filter(k -> keysetMap.containsKey(k.getKeysetId()) && keysetMap.get(k.getKeysetId()).getSiteId() != Data.RefreshKeySiteId)
```
jon8787, line 329: We shouldn't need to use RefreshKeySiteId since we know RefreshKey's keyset id

src/main/java/com/uid2/operator/Const.java
```diff
@@ -1,6 +1,8 @@
 package com.uid2.operator;
 
 public class Const extends com.uid2.shared.Const {
+    public static Object Data;
```
jon8787, line 4: what's this for?

src/main/java/com/uid2/operator/Main.java
```diff
@@ -214,7 +214,7 @@ private ICloudStorage wrapCloudStorageForOptOut(ICloudStorage cloudStorage) {
 
     private void run() throws Exception {
         Supplier<Verticle> operatorVerticleSupplier = () -> {
-            UIDOperatorVerticle verticle = new UIDOperatorVerticle(config, clientKeyProvider, keyStore, keyAclProvider, saltProvider, optOutStore, Clock.systemUTC(), _statsCollectorQueue);
+            UIDOperatorVerticle verticle = new UIDOperatorVerticle(config, clientKeyProvider, keysetKeyStore, keysetProvider, saltProvider, optOutStore, Clock.systemUTC(), _statsCollectorQueue);
```
jon8787, line 217: can we pass keyManager here instead?

src/main/java/com/uid2/operator/vertx/ClientVersionCapturingHandler.java
```diff
@@ -1,6 +1,6 @@
 package com.uid2.operator.vertx;
 
-import com.uid2.shared.Const;
+import com.uid2.operator.Const;
```
jon8787, line 3: why do we need this change?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -271,44 +264,55 @@ public void handleKeysRequest(RoutingContext rc) {
...
-                } else if (!acls.canClientAccessKey(clientKey, key, mode)) {
+            KeysetKey masterKey = this.keyManager.getActiveKeyBySiteId(Data.MasterKeySiteId, Instant.now());
+            if (masterKey == null) {
+                throw new RuntimeException(String.format("Cannot get active master key with SITE ID %d.", Data.MasterKeySiteId));
+            }
+
+            // defaultKeysetId allows calling sdk.Encrypt(rawUid) without specifying the keysetId
+            int defaultKeysetId = this.keyManager.getActiveKeyBySiteIdWithFallback(clientKey.getSiteId(), Data.AdvertisingTokenSiteId, Instant.now()).getKeysetId();
+
+            // include 'keyset_id' field, if:
```
jon8787, line 285: we don't want fallback here

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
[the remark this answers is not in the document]
jon8787: Can we write a test case that will discover if anything's wrong here?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -148,8 +157,37 @@ private static byte[] makeAesKey(String prefix) {
         return String.format("%1$16s", prefix).getBytes();
     }
```
jon8787, line 158: Do we really need all these mocks? Can we just mock RotatingKeysetProvider.getSnapshot() and IKeysetKeyStore.getSnapshot() instead?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -2344,6 +2463,554 @@ void requestWithReferer(Vertx vertx, VertxTestContext testContext) {
...
+            return snapshot.canClientAccessKey(i.getArgument(0, ClientKey.class), i.getArgument(1, KeysetKey.class), i.getArgument(2, MissingAclMode.class));
+        });
+
+        when(keysetKeyStoreSnapshot.getActiveKey(anyInt(), any())).then(i -> {
+            KeysetKeyStoreSnapshot snapshot = new KeysetKeyStoreSnapshot(new HashMap<>(keyMap), new HashMap<>(keysetMap));
+            return snapshot.getActiveKey(i.getArgument(0, Integer.class), i.getArgument(1, Instant.class));
+        });
+    }
+
+    void setKeysets(Map<Integer, Keyset> keysets) {
```
cody-constine-ttd, line 2728: Why did you go back on the way I setup the keysets and keysetKeys?

conf/default-config.json
zaiweidu: converted to Const.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -317,11 +322,13 @@ public void handleKeysSharing(RoutingContext rc) {
         }
     }
 
-    private List<EncryptionKey> getEncryptionKeys() {
-        final List<EncryptionKey> keyStore = this.keyStore.getSnapshot().getActiveKeySet()
-                .stream().filter(k -> k.getSiteId() != Const.Data.RefreshKeySiteId)
+    private List<KeysetKey> getKeysetKeys() {
```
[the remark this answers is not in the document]
zaiweidu: Fixed.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -271,44 +264,55 @@ public void handleKeysRequest(RoutingContext rc) {
...
                 }
-                keySet.put("id", key.getId());
-                keySet.put("created", key.getCreated().getEpochSecond());
-                keySet.put("activates", key.getActivates().getEpochSecond());
-                keySet.put("expires", key.getExpires().getEpochSecond());
-                keySet.put("secret", EncodingUtils.toBase64String(key.getKeyBytes()));
-
+                if(clientKey.getSiteId() == keyset.getSiteId() && keyset.isEnabled()) {
+                    keyObj.put("keyset_id", key.getKeysetId());
+                } else if (keyset.getSiteId() == Data.MasterKeySiteId && keyset.isEnabled()) {
```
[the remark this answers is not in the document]
zaiweidu: keys from client's disabled keysets won't return to client.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
zaiweidu: Removed fallback.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
zaiweidu: This's one of many historical tests that need a cleanup later. I'll create a Continous Improvement ticket to track the work.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
zaiweidu: It was conflicting with my other Mockito interceptions. I've refactored it to an embedded class to work for my DSP & Pub tests. We do need Continous Improvement ticket to clean up / refactor unit tests on Operator.

src/main/java/com/uid2/operator/Const.java
```diff
@@ -14,5 +14,8 @@ public class Config extends com.uid2.shared.Const.Config {
         public static final String FailureShutdownWaitHoursProp = "failure_shutdown_wait_hours";
         public static final String AllowLegacyAPIProp = "allow_legacy_api";
         public static final String SharingTokenExpiryProp = "sharing_token_expiry_seconds";
+        public static final int MasterKeyKeysetId = 1;
```
CodyConstine, line 17: We made these const in shared

zaiweidu: I was thinking the same thing. But I can't wait for the tiny thing. I'll include it in Continuous Improvement ticket.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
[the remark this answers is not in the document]
zaiweidu: ID_READER & GENERATOR test cases are added.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
zaiweidu: Logger added.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
zaiweidu: If we want to use the real logic of key lookup and sharing defined in Shared instead of mocks, we have to instantiate StoreSnapshot and ProviderSnapshot and make them cross referrenced by Keyset ID. This is exactly what I'm doing in _linkKeysToKeysets()_.

src/main/java/com/uid2/operator/Const.java
jon8787: @zaiweidu please see https://github.com/IABTechLab/uid2-shared/commit/691c76475f246a9587cccdcd4f1f96814a19cab3 , no need to wait

src/main/java/com/uid2/operator/Main.java
```diff
@@ -119,23 +120,22 @@ public Main(Vertx vertx, JsonObject config) throws Exception {
...
             this.clientKeyProvider.loadContent();
-            this.keyStore.loadContent();
-            this.keyAclProvider.loadContent();
             this.saltProvider.loadContent();
+            this.keysetProvider.loadContent();
+            this.keysetKeyStore.loadContent();
         }
-
-        metrics = new OperatorMetrics(keyStore, saltProvider);
+        this.keyManager = new KeyManager(this.keysetKeyStore, this.keysetProvider);
```
jon8787, line 137: Doesn't seem right that we store keyManager as a member but then re-create it on line 217. How about we remove the member variable and replace with a getKeyManager() function?

src/main/java/com/uid2/operator/Const.java
zaiweidu: picked it up.

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
```diff
@@ -209,8 +208,8 @@ public AdvertisingToken decodeAdvertisingTokenV2(Buffer b) {
                     TokenVersion.V2,
                     Instant.ofEpochMilli(establishedMillis),
                     Instant.ofEpochMilli(expiresMillis),
-                    new OperatorIdentity(0, OperatorType.Service, 0, 0),
-                    new PublisherIdentity(siteId, 0, 0),
+                    new OperatorIdentity(0, OperatorType.Service, 0, masterKeyId),
```
jon8787, line 211: why do we need these changes?

zaiweidu: @cody-constine-ttd made the specific change.

src/main/java/com/uid2/operator/Const.java
jon8787: I don't see it.

src/main/java/com/uid2/operator/service/RoutingContextReader.java
```diff
@@ -1,6 +1,6 @@
 package com.uid2.operator.service;
 
-import com.uid2.shared.Const;
+import com.uid2.operator.Const;
```
jon8787, line 3: can we revert this?

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
jon8787: let's find out

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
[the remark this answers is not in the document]
jon8787: how? this "if" statement is only about keyset_id, not about other fields in the key

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -271,44 +272,52 @@ public void handleKeysRequest(RoutingContext rc) {
...
+                    keyObj.put("keyset_id", key.getKeysetId());
+                } else if (!this.keyManager.canClientAccessKey(clientKey, key, mode)) {
```
jon8787, line 304: what will happen if getSnapshot() returns a new snapshot inside this loop?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1416,11 +1418,15 @@ private JsonObject toJsonV1(IdentityTokens t) {
...
+        MissingAclMode mode = MissingAclMode.DENY_ALL;
+        if (clientKey.getRoles().contains(Role.ID_READER)) {
+            mode = MissingAclMode.ALLOW_ALL;
+        }
         final JsonArray a = new JsonArray();
         for (int i = 0; i < keys.size(); ++i) {
-            final EncryptionKey k = keys.get(i);
-            if (!acls.canClientAccessKey(clientKey, k)) {
+            final KeysetKey k = keys.get(i);
+            if (!this.keyManager.canClientAccessKey(clientKey, k, mode)) {
```
jon8787, line 1429: what happens if the getSnapshot() call returns a new snapshot inside this loop?

src/test/java/com/uid2/operator/TokenEncodingTest.java
```diff
@@ -23,23 +26,33 @@
 
 public class TokenEncodingTest {
 
-    private final IKeyStore keyStoreInstance;
+    private final IKeysetKeyStore keysetKeyStoreInstance;
+    private final RotatingKeysetProvider keysetProviderInstance;
```
jon8787, line 31: should we be using KeyManager as a member instead of these 2?

src/main/java/com/uid2/operator/model/KeyManager.java
jon8787: I don't see it. Which unit test depends on this method being public?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -2344,6 +2463,537 @@ void requestWithReferer(Vertx vertx, VertxTestContext testContext) {
                 });
     }
 
+    /**********************************************
+     * SHARING TESTS
```
jon8787, line 2467: These are DSP tests too?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -2344,6 +2463,537 @@ void requestWithReferer(Vertx vertx, VertxTestContext testContext) {
...
+                KeysetKey key129 = new KeysetKey(1129, makeAesKey("key129"), now.minusSeconds(10), now.minusSeconds(5), now.minusSeconds(5), 11);
+                test.AddKeyset(11, keyset11);
+                test.AddKey(1128, key128);
+                test.AddKey(1129, key129); // key129 is expired but should return
+                resetMocks();
+                test.setupMockitoApiInterception();
+                expectedKeys.add(key128);
+                expectedKeys.add(key129);
+                break;
+
```
jon8787, line 2941: - Instead of test1, test2, can we name these to be human readable? eg "SameSiteId", "MissingAllowedSites", "AllowedSitesIncludeUs"
- Even better if we could make these separate tests and use common functions to do the setup and asserts

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
zaiweidu: Take a snapshot on keysets out side of loop.

src/main/java/com/uid2/operator/Main.java
zaiweidu: Replaced member this.keyManager with getKeyManager() method.

src/test/java/com/uid2/operator/TokenEncodingTest.java
zaiweidu: Fixed as recommended.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
zaiweidu: DSPs & Pubs are down below.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
zaiweidu: Cached snapshot and keysets outside of loop.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
zaiweidu: Cached a keysetSnapshot outside of loop and use inside.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
zaiweidu: Applied hardcoded keyset IDs for master/refresh/publisher. The values are defined in Shared.

src/main/java/com/uid2/operator/model/KeyManager.java
zaiweidu: Refactor code and make it private.

src/main/java/com/uid2/operator/model/KeyManager.java
zaiweidu: Logs added before throw.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
jon8787: I still see "test1", "test2", "test3", "test4"

zaiweidu: I didn't push... avoid disturbing your review. I renamed the test cases with meaningful names in the latest push.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
zaiweidu: I have refactored KeyManager class to support Snapshot mode. KeyManager::Snapshot() will cache keys/keysets within its lifetime. This solves the consistency concern here and the other location.

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
zaiweidu: This change was made to test & validate 'token/generate' in the new test case tokenGenerateRotatingKeysets_GENERATOR()

jon8787: Ok got it. decodeAdvertisingToken() is only used by /token/validate and tests. /token/validate doesn't use these values, and tests never consumed these values until now.

zaiweidu (commented):
Please review.

src/test/java/com/uid2/operator/RotatingKeysetKeyStoreTest.java
jon8787: Could you explain a bit further? Why are we testing it here and not in shared?

src/main/java/com/uid2/operator/model/KeyManager.java
zaiweidu: I've removed the log as we throw the error.

src/test/java/com/uid2/operator/RotatingKeysetKeyStoreTest.java
zaiweidu: These tests don't test functionality of rotating key store or provider, but The ensure existence of test json (filenames & paths) for Operator unit tests. This can't be done in Shared.

zaiweidu (commented):
@jon8787 Please review.

src/main/java/com/uid2/operator/monitoring/OperatorMetrics.java
zaiweidu: Per our discussion today, let's only monitor active keys from default keysets for now. I've create UID2-1556 as Continuous Improvement to track the future work.

zaiweidu (commented):
will do more local integration tests.

jon8787 (changes requested):

PR #154, opened 2023-08-03
UID2-934 refresh token before expire

cYKatherine:


src/main/java/com/uid2/operator/Main.java
```diff
@@ -398,30 +399,32 @@ public DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticC
...
-            return new UidCoreClient(attestationUrl, userToken, this.appVersion, CloudUtils.defaultProxy, AttestationFactory.getGcpOidcAttestation(), enforceHttps);
-        }
-        else if(enclavePlatform != null && enclavePlatform.equals("azure-sgx"))
-        {
-            LOGGER.info("creating uid core client with azure sgx attestation protocol");
-            return new UidCoreClient(attestationUrl, userToken, this.appVersion, CloudUtils.defaultProxy, AttestationFactory.getAzureAttestation(), enforceHttps);
+        AttestationTokenRetriever attestationTokenRetriever;
+
+        switch (enclavePlatform) {
+            default:
```
thomasm-ttd, line 408: I would always put the default last, as that is what gets executed if nothing else matches. Makes it easier to read

thomasm-ttd (approved):

thomasm-ttd (approved):

PR #160, opened 2023-08-07
Syw UID2-1224 special email opt out handling

sunnywu:
1.  Always enforce the special identities optout/optin status regardless what the optout store would state. 
2. This enforcement applied to token/generate, token/client-generate and token/refresh
3. All optout check goes thru UID2OperatorService.hasGlobalOptOut so the special identities like above are handled in this method now

### DIFF
--- a/src/main/java/com/uid2/operator/service/UIDOperatorService.java
+++ b/src/main/java/com/uid2/operator/service/UIDOperatorService.java
@@ -32,10 +32,6 @@ public class UIDOperatorService implements IUIDOperatorService {
     private final IdentityScope identityScope;
     private final UserIdentity testOptOutIdentityForEmail;
     private final UserIdentity testOptOutIdentityForPhone;
-
-    private final UserIdentity testAlwaysOptInIdentityForEmail;
-    private final UserIdentity testAlwaysOptInIdentityForPhone;
-
     private final Duration identityExpiresAfter;
     private final Duration refreshExpiresAfter;
     private final Duration refreshIdentityAfter;
@@ -56,10 +52,6 @@ public class UIDOperatorService implements IUIDOperatorService {
                 InputUtil.normalizeEmail("optout@email.com").getIdentityInput(), Instant.now());
         this.testOptOutIdentityForPhone = getFirstLevelHashIdentity(identityScope, IdentityType.Phone,
                 InputUtil.normalizePhone("+00000000000").getIdentityInput(), Instant.now());
-        this.testAlwaysOptInIdentityForEmail = getFirstLevelHashIdentity(identityScope, IdentityType.Email,
-                InputUtil.normalizeEmail("valid-token@email.com").getIdentityInput(), Instant.now());
-        this.testAlwaysOptInIdentityForPhone = getFirstLevelHashIdentity(identityScope, IdentityType.Phone,
-                InputUtil.normalizePhone("+00000000001").getIdentityInput(), Instant.now());
 
         this.operatorIdentity = new OperatorIdentity(0, OperatorType.Service, 0, 0);
 
@@ -247,9 +239,6 @@ public class UIDOperatorService implements IUIDOperatorService {
     private Tuple.Tuple2<Boolean, Instant> hasGlobalOptOut(UserIdentity userIdentity) {
         if (userIdentity.matches(testOptOutIdentityForEmail) || userIdentity.matches(testOptOutIdentityForPhone)) {
             return new Tuple.Tuple2<>(true, Instant.now());
-        }
-        else if (userIdentity.matches(testAlwaysOptInIdentityForEmail) || userIdentity.matches(testAlwaysOptInIdentityForPhone)) {
-            return new Tuple.Tuple2<>(false, null);
         }
         Instant result = this.optOutStore.getLatestEntry(userIdentity);
         return new Tuple.Tuple2<>(result != null, result);
--- a/src/main/java/com/uid2/operator/service/UIDOperatorService.java
+++ b/src/main/java/com/uid2/operator/service/UIDOperatorService.java
@@ -85,7 +85,7 @@ public class UIDOperatorService implements IUIDOperatorService {
                 request.userIdentity.identityScope, request.userIdentity.identityType, firstLevelHash, request.userIdentity.privacyBits,
                 request.userIdentity.establishedAt, request.userIdentity.refreshedAt);
 
-        if (request.shouldCheckOptOut() && hasGlobalOptOut(firstLevelHashIdentity).isOptedOut()) {
+        if (request.shouldCheckOptOut() && getGlobalOptOut(firstLevelHashIdentity).isOptedOut()) {
             return IdentityTokens.LogoutToken;
         } else {
             return generateIdentity(request.publisherIdentity, firstLevelHashIdentity);
@@ -108,7 +108,7 @@ public class UIDOperatorService implements IUIDOperatorService {
         }
 
         try {
-            final GlobalOptoutResult logoutEntry = hasGlobalOptOut(token.userIdentity);
+            final GlobalOptoutResult logoutEntry = getGlobalOptOut(token.userIdentity);
             boolean optedOut = logoutEntry.isOptedOut();
 
             if (!optedOut || token.userIdentity.establishedAt.isAfter(logoutEntry.getTime())) {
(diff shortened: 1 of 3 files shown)


src/main/java/com/uid2/operator/util/Tuple.java
```diff
@@ -6,9 +6,6 @@ public static class Tuple2<T1, T2> {
         private final T2 item2;
 
         public Tuple2(T1 item1, T2 item2) {
-            assert item1 != null;
-            assert item2 != null;
```
sunnywu, line 10: i removed these assert as my hasGlobalOptOut function above can legit return null item. is this ok @vishalegbert-ttd ?

vishalegbert-ttd: Should be alright

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -86,7 +86,7 @@ public IdentityTokens generateIdentity(IdentityRequest request) {
                 request.userIdentity.identityScope, request.userIdentity.identityType, firstLevelHash, request.userIdentity.privacyBits,
                 request.userIdentity.establishedAt, request.userIdentity.refreshedAt);
 
-        if (request.shouldCheckOptOut() && hasGlobalOptOut(firstLevelHashIdentity)) {
+        if (request.shouldCheckOptOut() && hasGlobalOptOut(firstLevelHashIdentity).getItem1()) {
```
jon8787, line 89: in future let's try to use a class (even if it has public members) instead of tuple, as .getItem1() is harder to read than say .isOptedOut

src/test/java/com/uid2/operator/UIDOperatorServiceTest.java
```diff
@@ -254,4 +257,106 @@ public void testIdentityMapForOptOutUser() {
...
+
+    //UID2 uses v2 tokens but v2 doesn't handle phone number IdentityType correctly
+    //so passing in a phone number and it will be still casted as an email type in UserIdentity
+    //and UserIdentity won't match to the default UIDOperatorService.testAlwaysOptInIdentityForPhone
+    //as the IdentityType won't match
+    //will only test when we switch to v4 token
+    //this works for EUID because EUID is on v3 token already which persists to correct IdentityType
+    @ParameterizedTest
+    @CsvSource({"email,optout@email.com,UID2", "email,optout@email.com,EUID", "phone,+00000000000,EUID"})
+    public void testSpecialIdentityOptOutTokenRefresh(String type, String id, IdentityScope scope) {
```
jon8787, line 294: this is not testing your changes right? this is a test that should have already been there?

src/test/java/com/uid2/operator/UIDOperatorServiceTest.java
```diff
@@ -254,4 +257,106 @@ public void testIdentityMapForOptOutUser() {
...
+        else
+        {
+            inputVal = InputUtil.normalizePhone(id);
+        }
+        //make sure this still works without optout record
+        when(this.optOutStore.getLatestEntry(any())).thenReturn(null);
+
+        final IdentityRequest identityRequest = new IdentityRequest(
+                new PublisherIdentity(123, 124, 125),
+                inputVal.toUserIdentity(IdentityScope.UID2, 0, this.now),
```
jon8787, line 279: why do we not need EUID here?

src/test/java/com/uid2/operator/UIDOperatorServiceTest.java
```diff
@@ -254,4 +257,106 @@ public void testIdentityMapForOptOutUser() {
...
+                inputVal.toUserIdentity(scope, 0, this.now),
+                TokenGeneratePolicy.JustGenerate
+        );
+        IdentityTokens tokens;
+        if(scope == IdentityScope.EUID) {
+            tokens = euidService.generateIdentity(identityRequest);
+        }
+        else {
+            tokens = uid2Service.generateIdentity(identityRequest);
+        }
```
jon8787, line 349: a lot of this code looks similar to testSpecialIdentityOptOutTokenGenerate, can we re-use?

jon8787 (approved):

src/test/java/com/uid2/operator/UIDOperatorServiceTest.java
sunnywu: fair point - i wrote the test below last and forgot to modify this older test... fixed

src/test/java/com/uid2/operator/UIDOperatorServiceTest.java
sunnywu: yes but i just wanna test thoroughly with EUID variant

src/test/java/com/uid2/operator/UIDOperatorServiceTest.java
sunnywu: i thought about that but i think it will be too messy to add another variance of expectation (expectation Optout token versus non optout tokens) so i prefer to keep it this way

jon8787: Right, I just meant the beginning of the method could be a private method that both tests could call

jon8787 (approved):

src/test/java/com/uid2/operator/UIDOperatorServiceTest.java
sunnywu: ah right will just leave it as is for now :)

src/test/java/com/uid2/operator/UIDOperatorServiceTest.java
```diff
@@ -254,4 +257,112 @@ public void testIdentityMapForOptOutUser() {
         assertNotNull(mappedIdentityShouldBeOptOut);
         assertTrue(mappedIdentityShouldBeOptOut.isOptedOut());
     }
+
+    //UID2-1224
+    @ParameterizedTest
+    @CsvSource({"email,optout@email.com,UID2", "email,optout@email.com,EUID", "phone,+00000000000,EUID"})
```
mcollins-ttd, line 263: Should we have tests for the hashes of these test identities?

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -239,8 +236,33 @@ private AdvertisingToken createAdvertisingToken(PublisherIdentity publisherIdent
                 userIdentity);
     }
 
-    private boolean hasGlobalOptOut(UserIdentity userIdentity) {
-        return this.optOutStore.getLatestEntry(userIdentity) != null;
+    protected class GlobalOptoutResult {
```
mcollins-ttd, line 239: Am I right thinking that `time == null` if and only if `isOptedOut == true`?

If so I suggest either:

1. Change the ctor from `GlobalOptoutResult(boolean providedIsOptedOut, Instant providedTime)` to `GlobalOptoutResult(Instant time)` and make `isOptedOut()` return `time != null`, or
2. Rename `hasGlobalOptout` to something like `getOptoutTime` and have it return an `Instant`. Consumer code can then just check the returned value against null to see if the identity is opted out.

jon8787: option 1

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -1,6 +1,7 @@
 package com.uid2.operator.service;
 
 import com.uid2.operator.model.*;
+import com.uid2.operator.util.Tuple;
```
mcollins-ttd, line 4: This can be removed

jon8787 (approved):

Ian-Nara:
LGTM

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -86,7 +85,7 @@ public IdentityTokens generateIdentity(IdentityRequest request) {
                 request.userIdentity.identityScope, request.userIdentity.identityType, firstLevelHash, request.userIdentity.privacyBits,
                 request.userIdentity.establishedAt, request.userIdentity.refreshedAt);
 
-        if (request.shouldCheckOptOut() && hasGlobalOptOut(firstLevelHashIdentity)) {
+        if (request.shouldCheckOptOut() && hasGlobalOptOut(firstLevelHashIdentity).isOptedOut()) {
```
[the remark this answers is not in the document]
sunnywu: renamed to `getGlobalOptoutResult`

mcollins-ttd (approved):

sunnywu:
Created a new identical PR to merge to master in https://github.com/IABTechLab/uid2-operator/pull/176

PR #163, opened 2023-08-09
Get JWT from attest response and pass to Core and Optout

thomasm-ttd:


src/main/java/com/uid2/operator/Main.java
```diff
@@ -401,32 +400,31 @@ public DistributionStatisticConfig configure(Meter.Id id, DistributionStatisticC
 
     private UidCoreClient createUidCoreClient(String attestationUrl, String userToken, Handler<Integer> responseWatcher) throws Exception {
         String enclavePlatform = this.config.getString("enclave_platform", "");
+        if (enclavePlatform == null) {
```
aulme, line 403: This is already defaulted in the line above

thomasm-ttd: Should be, but we had it failing yesterday. Couldn't understand why and don't want to revert it

aulme (approved):

PR #167, opened 2023-08-09
UID2-1543 Operator support for keysets 

jon8787:
This PR enables Operator to:

- refresh keysets from Core (and from local for testing purposes)
- refresh keyset_keys from Core, replacing the /key/refresh call.
- no longer refresh legacy keys from Core
- change /key/sharing endpoint to respond with the actual default keyset and master keyset (instead of the hardcoded 99999 keyset which was used when Sharing used DEP ACLs)
- change all references to legacy key usage (keys.json) to keysets.json and keyset_keys.json (mostly via the KeyManager class)
- add tests [including those in this spec](https://atlassian.thetradedesk.com/confluence/display/UID2/UID2-545+Sharing+-+Data+Storage+for+Keys+and+Grants#UID2545SharingDataStorageforKeysandGrants-TestcasesforKeysets(July2023))


src/main/java/com/uid2/operator/service/RoutingContextReader.java
```diff
@@ -1,6 +1,6 @@
 package com.uid2.operator.service;
 
-import com.uid2.shared.Const;
+import com.uid2.operator.Const;
```
cody-constine-ttd, line 3: This isn't going to cause issues changing form the shared const to the operator const?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1416,11 +1414,19 @@ private JsonObject toJsonV1(IdentityTokens t) {
...
     }
 
-    private JsonArray toJson(List<EncryptionKey> keys, ClientKey clientKey, IKeysAclSnapshot acls) {
+    private JsonArray toJson(List<KeysetKey> keys, ClientKey clientKey) {
+        MissingAclMode mode = MissingAclMode.DENY_ALL;
+        if (clientKey.getRoles().contains(Role.ID_READER)) {
+            mode = MissingAclMode.ALLOW_ALL;
+        }
+
+        KeyManagerSnapshot keyManagerSnapshot = this.keyManager.getKeyManagerSnapshot(clientKey.getSiteId());
```
cody-constine-ttd, line 1423: I know this is from the old operator code, but we probably shouldn't be handling this kind of logic in a toJson function

src/main/java/com/uid2/operator/model/KeyManager.java
```diff
@@ -0,0 +1,126 @@
...
+        KeysetKey key = getActiveKeyBySiteId(siteId, asOf);
+        if (key == null) key = getActiveKeyBySiteId(fallbackSiteId, asOf);
+        if (key == null) {
+            throw new IllegalArgumentException(String.format("Cannot get active key in default keyset with SITE ID %d or %d.", siteId, fallbackSiteId));
+        }
+        return key;
+    }
+
+    private Keyset getDefaultKeysetBySiteId(int siteId) {
+        List<Keyset> keysets = this.keysetProvider.getSnapshot().getAllKeysets().values().stream()
```
cody-constine-ttd, line 47: We are going to do this with every call to keys/sharing, is that going to be ok performance wise?

src/test/java/com/uid2/operator/RotatingKeysetKeyStoreTest.java
```diff
@@ -0,0 +1,21 @@
...
+
+public class RotatingKeysetKeyStoreTest {
+    @Test
+    public void loadFromEmbeddedResourceStorage() throws Exception {
+        RotatingKeysetKeyStore keysetKeyStore = new RotatingKeysetKeyStore(
+                new EmbeddedResourceStorage(Main.class),
+                new GlobalScope(new CloudPath("/com.uid2.core/test/keyset_keys/metadata.json")));
+
+        JsonObject m = keysetKeyStore.getMetadata();
+        keysetKeyStore.loadContent(m);
```
cody-constine-ttd, line 19: Are we just testing no exception is thrown here? LoadContent is tested in shared

src/main/java/com/uid2/operator/model/KeyManager.java
```diff
@@ -0,0 +1,126 @@
...
+                this.keysetProvider.getSnapshot(),
+                this.getAllKeysets(),
+                this.getKeysetKeys(),
+                this.getMasterKey(),
+                this.getDefaultKeysetBySiteId(siteId));
+    }
+
+    public KeysetKey getActiveKeyBySiteIdWithFallback(int siteId, int fallbackSiteId, Instant asOf) {
+        KeysetKey key = getActiveKeyBySiteId(siteId, asOf);
+        if (key == null) key = getActiveKeyBySiteId(fallbackSiteId, asOf);
```
atarassov-ttd, line 39: As a followup, can we please add some monitoring for this. I'd like to kill the fallback logic off.

src/main/java/com/uid2/operator/model/KeyManager.java
```diff
@@ -0,0 +1,126 @@
...
+
+    private Keyset getDefaultKeysetBySiteId(int siteId) {
+        List<Keyset> keysets = this.keysetProvider.getSnapshot().getAllKeysets().values().stream()
+                .filter(s -> s.isEnabled() && s.isDefault() && s.getSiteId() == siteId).collect(Collectors.toList());
+        if (keysets.isEmpty()) return null;
+        return keysets.get(0);
+    }
+
+    // Retrieve an active key from default keyset by caller's site id.
+    public KeysetKey getActiveKeyBySiteId(int siteId, Instant asOf) {
```
atarassov-ttd, line 54: Similar to Cody's question above, this is probably going to be called on every token/generate and refresh -- should the data be pre-processed to optimise lookups by site id?

atarassov-ttd: If there is no immediate substantial perf impact, happy for this to be done as a separate PR.

src/main/java/com/uid2/operator/model/KeyManager.java
```diff
@@ -0,0 +1,126 @@
...
+    private KeysetKey getActiveKey(int keysetId, Instant asOf) {
+        return this.keysetKeyStore.getSnapshot().getActiveKey(keysetId, asOf);
+    }
+
+    public KeysetKey getKey(int keyId) {
+        return this.keysetKeyStore.getSnapshot().getKey(keyId);
+    }
+
+    private List<KeysetKey> getActiveKeysetKeys() {
+        // return all keys without expiry check
```
atarassov-ttd, line 77: What does this mean? Aren't active keys by definition not expired?

src/main/java/com/uid2/operator/model/KeyManager.java
```diff
@@ -0,0 +1,126 @@
...
+        return this.keysetKeyStore.getSnapshot().getActiveKeysetKeys();
+    }
+
+    public List<KeysetKey> getKeysetKeys() {
+        Map<Integer, Keyset> keysetMap = this.keysetProvider.getSnapshot().getAllKeysets();
+        List<KeysetKey> keys = getActiveKeysetKeys();
+        return keys
+                .stream().filter(k -> keysetMap.containsKey(k.getKeysetId()) && k.getKeysetId() != Const.Data.RefreshKeysetId)
```
atarassov-ttd, line 85: Does filtering out refresh keys really belong here? Sounds like logic that belongs to a higher level. It's not obvious from method name that refresh keyset would be excluded. If you want to keep the logic here, maybe change name to something like `getSharingKeys()`?

src/main/java/com/uid2/operator/model/KeyManagerSnapshot.java
```diff
@@ -0,0 +1,44 @@
...
+import com.uid2.shared.auth.Keyset;
+import com.uid2.shared.auth.KeysetSnapshot;
+import com.uid2.shared.model.KeysetKey;
+
+import java.util.List;
+import java.util.Map;
+
+public class KeyManagerSnapshot {
+    private final KeysetSnapshot keysetSnapshot;
+    private final Map<Integer, Keyset> keysetMap;
```
atarassov-ttd, line 12: It is not immediately obvious what the key is here. A common or more explicit name could help here.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -271,59 +266,62 @@ public void handleKeysRequest(RoutingContext rc) {
...
-                keySet.put("activates", key.getActivates().getEpochSecond());
-                keySet.put("expires", key.getExpires().getEpochSecond());
-                keySet.put("secret", EncodingUtils.toBase64String(key.getKeyBytes()));
+            final JsonObject resp = new JsonObject();
+            resp.put("caller_site_id", clientKey.getSiteId());
+            resp.put("master_keyset_id", masterKey.getKeysetId());
+            if (defaultKeyset != null) {
+                resp.put("default_keyset_id", defaultKeyset.getKeysetId());
+            }
+            resp.put("token_expiry_seconds", this.config.getString(Const.Config.SharingTokenExpiryProp));
```
atarassov-ttd, line 292: Should this config be extracted on startup and cached? What if the property not available in the config?

src/main/java/com/uid2/operator/service/RoutingContextReader.java
jon8787: reverting this

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
jon8787: will rename as getAccessibleKeysAsJson (I'm not sure that we need a vanilla toJson method yet)

src/main/java/com/uid2/operator/model/KeyManager.java
jon8787: [UID2-1631](https://atlassian.thetradedesk.com/jira/browse/UID2-1631)

src/main/java/com/uid2/operator/model/KeyManagerSnapshot.java
jon8787: Agree, I intended to make this comment but there were too many rounds of review. Renaming to keysetIdToKeyset

src/test/java/com/uid2/operator/RotatingKeysetKeyStoreTest.java
jon8787: Don't know much about why this was added. Is this test different because it's testing the embedded resource? 
Why is RotatingKeysetProviderTest in Operator rather than shared?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
jon8787: [this is already in prod](https://github.com/IABTechLab/uid2-operator/blob/f255c13252922a4ccdb8a0fc95077e40459dd1bb/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java#L279), O(1) map lookup.

We could add a default if not available but as it's in prod, prefer to leave it for now

src/main/java/com/uid2/operator/model/KeyManager.java
jon8787: Removing this comment. @cody-constine-ttd why are these called "active" keys? Expired keys are included in this list, (they're needed by /key/sharing requesters for decryption)

jon8787: Ok @cody-constine-ttd said it's to differentiate these from the 30 day old keys. But those keys are deleted, so I'm not sure I agree with the "active" terminology which includes expired keys but not deleted keys.

jon8787: this led me down a rabbit hole, researching key deletion etc. Conclusion is that getActiveKeysetKeys() is incorrectly named both here and in uid2-shared. Will fix it here and can fix it in shared later. The other "active" methods are correct.

src/main/java/com/uid2/operator/model/KeyManager.java
jon8787: moving it is a bit painful. It's also used by DSPs. Will go with `getKeysForSharingOrDsps()` if no concerns

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
atarassov-ttd: Not that concerned about performance. More about config lookup logic being sprinkled all over place. If we add a default, how would we ensure that this lookup/default logic does not get copy-pasted somewhere?

jon8787: How do you prevent copy/paste in general... taken to extremes we should not allow "config" to be a member variable. 
How about if I replace this with:

    private String getSharingTokenExpirySeconds() {
        return config.getString(Const.Config.SharingTokenExpiryProp);
    }

atarassov-ttd: That's definitely better. I also like processing config into (final) members because it allows to validate the config at startup instead of waiting until somebody happens to hit the endpoint. Given that eventually this sort of settings should be managed centrally, having (final) members may not work. Happy with the proposed resolution for now.

src/main/java/com/uid2/operator/model/KeyManager.java
jon8787: [UID2-1634](https://atlassian.thetradedesk.com/jira/browse/UID2-1634)

src/main/java/com/uid2/operator/model/KeyManager.java
jon8787: I'm not happy with it either. In the interest of closing this PR sooner, raised [UID2-1634](https://atlassian.thetradedesk.com/jira/browse/UID2-1634)

sunnywu: can this be done (in UID2-1634 maybe) everytime a keyset/keysetkey refresh completed and there's a change to existing copies? so just do it for every site once after a refresh is complete.

jon8787: We can, but we're doing a linear search anyway to find accessible keys (canClientAccessKey). Number of keysets will only be ~200 in prod so it's going to be fast for a while longer.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -2360,6 +2285,601 @@ void requestWithReferer(Vertx vertx, VertxTestContext testContext) {
...
+                new KeysetKey(1004, makeAesKey("key4"), now.minusSeconds(10), now.minusSeconds(6), now.plusSeconds(3600), 4),
+                new KeysetKey(1005, makeAesKey("key5"), now.minusSeconds(10), now.minusSeconds(4), now.plusSeconds(3600), 4),
+                new KeysetKey(1006, makeAesKey("key6"), now.minusSeconds(10), now.minusSeconds(2), now.plusSeconds(3600), 4),
+                new KeysetKey(1007, makeAesKey("key7"), now.minusSeconds(10), now, now.plusSeconds(3600), 4),
+                new KeysetKey(1008, makeAesKey("key8"), now.minusSeconds(10), now.plusSeconds(5), now.plusSeconds(3600), 4),
+                new KeysetKey(1009, makeAesKey("key9"), now.minusSeconds(10), now.minusSeconds(5), now.minusSeconds(2), 4),
+                // keys in keyset5
+                new KeysetKey(1010, makeAesKey("key10"), now.minusSeconds(10), now, now.plusSeconds(3600), 5),
+                new KeysetKey(1011, makeAesKey("key11"), now.minusSeconds(10), now.plusSeconds(5), now.plusSeconds(3600), 5),
+                new KeysetKey(1012, makeAesKey("key12"), now.minusSeconds(10), now.minusSeconds(5), now.minusSeconds(2), 5),
```
sunnywu, line 2592: mention why keyset6 (disabled) and keyset8 (not sharing with site id 101)  as comments for better tracability

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -2360,6 +2285,601 @@ void requestWithReferer(Vertx vertx, VertxTestContext testContext) {
...
```
sunnywu, line 2656: mention why keyset6/7/8 as comments for better tracability

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -2360,6 +2285,601 @@ void requestWithReferer(Vertx vertx, VertxTestContext testContext) {
...
+                test.addKey(1232, key232);
+                test.setupMockitoApiInterception();
+                expectedKeys.add(key229);
+                expectedKeys.add(key230);
+                expectedKeys.add(key232);
+                break;
+
+            // c. Rotate keys within a keyset. Test /key/sharing shows the rotation
+            case "RotateKey":
+                KeysetKey key329 = new KeysetKey(1329, makeAesKey("key229"), now.minusSeconds(10), now /* activate immediately */, now.plusSeconds(3600), 4); // default keyset
```
sunnywu, line 2850: _makeAesKey("key229")_ 

key329?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -2360,6 +2285,601 @@ void requestWithReferer(Vertx vertx, VertxTestContext testContext) {
...
+                test.addKey(1329, key329);
+                test.setupMockitoApiInterception();
+                expectedKeys.add(key329);
+                KeysetKey actual = keysetKeyStore.getSnapshot().getActiveKey(4, Instant.now());
+                assertEquals(key329, actual);
```
sunnywu, line 2855: why is it key329 not key1007 ?

sunnywu: don't they activate at the same time ("now")?

cody-constine-ttd (approved):
LGTM

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
jon8787: As discussed, getActiveKey() returns the last key that is active (meaning "activates" is not in the future, and "expires" is in the future)

I've added getActiveKeyTest() which is a more detailed test of this

sunnywu (approved):

PR #183, opened 2023-08-21
CSTG

sunnywu:
- basic CSTG functionality with feature switch and a single set of hardcoded client side keypair configs in config file


conf/local-config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "service_instances": 1,
```
sunnywu, line 2: should undo all the non-cstg changes in this file

mcollins-ttd (approved):

PR #193, opened 2023-08-31
Syw UId2-1354 key sharing endpoint providing site domain name data

sunnywu:
1. providing site domain names in the v2/key/sharing endpoint 
2. Added key_sharing_endpoint_provide_site_domain_names feature switch so we can turn it off optionally (without needing to disable cstg feature switch)
3. Added more site domain name expectations in various key sharing unit tests
4. The extra output inside "body" of v2/key/sharing response will be like this:

The end result will look something like this:
```
                "site_data": [
                        {
                            "id": 101,
                            "domain_names": [
                                "101.co.uk",
                                "101.com"
                            ]
                        },
                        {
                            "id": 102,
                            "domain_names": [
                                "102.co.uk",
                                "102.com"
                            ]
                        }
                    ]
```


src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -3445,15 +3446,51 @@ void keySharingKeysets_CorrectFiltering(Vertx vertx, VertxTestContext testContex
         });
     }
 
-    @Test
+    //set some default domain names for all possible sites for each unit test first
+    private void setupSiteDomainNameListCall(int... siteIds) {
+        for(int siteId : siteIds) {
+            doReturn(new Site(siteId, "site"+siteId, true, new HashSet<>(Arrays.asList(siteId+".com", siteId+".co.uk")))).when(siteProvider).getSite(siteId);
```
jon8787, line 3452: Can we mock at a higher level? getSnapshot() rather than getSite() ?
eg see setupKeysetsMock()

(Eg, ISiteStore.getAllSites() won't work with the above code. Mocking getSnapshot() would make current and future ISiteStore methods work correctly)

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -599,9 +604,45 @@ public void handleKeysSharing(RoutingContext rc) {
...
+            // key/sharing endpoint
+            if(keySharingEndpointProvideSiteDomainNames && clientSideTokenGenerate) {
+                for (Integer siteId : accessibleSites) {
+                    Site s = siteProvider.getSite(siteId);
+                    if(s == null) {
+                        continue;
+                    }
+                    JsonObject siteObj = new JsonObject();
+                    siteObj.put("id", siteId);
+                    siteObj.put("domain_names", s.getDomainNames());
```
mcollins-ttd, line 622: If `s.getDomainNames()` is null or empty, can we omit that site from the response in order to reduce response size?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -599,9 +604,45 @@ public void handleKeysSharing(RoutingContext rc) {
...
+                accessibleSites.add(keyset.getSiteId());
             }
             resp.put("keys", keys);
-
+            //without cstg enabled, operator won't have site data and siteProvider could be null
+            //and adding keySharingEndpointProvideSiteDomainNames in case something goes wrong
+            //and we can still enable cstg feature but turn off site domain name download in
+            // key/sharing endpoint
+            if(keySharingEndpointProvideSiteDomainNames && clientSideTokenGenerate) {
+                for (Integer siteId : accessibleSites) {
```
mcollins-ttd, line 615: Note: validator cares about order of arrays, so it might complain about mismatches.

jon8787: Yes we may want to sort these so they're always in a deterministic order

src/test/java/com/uid2/operator/ExtendedUIDOperatorVerticle.java
```diff
@@ -28,4 +28,9 @@ public ExtendedUIDOperatorVerticle(JsonObject config,
     public IUIDOperatorService getIdService() {
         return this.idService;
     }
+
+    public void setKeySharingEndpointProvideSiteDomainNames(boolean enable) {
```
jon8787, line 32: Was discussing with @mcollins-ttd that there might be another way to do this that doesn't require inheriting from the real verticle, changing config after construction, and writing a method for every config we need to change. 

We could have another test class that doesn't create UIDOperatorVerticle in @BeforeEach deployVerticle() and instead relies on the @Test methods to do it. The test methods would call a function like this (this is @mcollins-ttd 's)

Future<String> Run(Vertx vertx) {
	//...

	return vertx.deployVerticle(uidOperatorVerticle);
}

@Test
void Foo(Vertx vertx, VertxTestContext testContext) {
	config.put("foo", "bar");
	Run(vertx).onComplete(testContext.succeeding(id -> {
		testContext.completeNow();
	}));
}

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
sunnywu: @jon8787 fixed please have a look if it looks like as you thought

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
sunnywu: Fixed as suggested

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
sunnywu: sorted the sites and also the domain name list in each site

src/test/java/com/uid2/operator/ExtendedUIDOperatorVerticle.java
sunnywu: as discussed previously, will do this improvement in a CI ticker [here](https://atlassian.thetradedesk.com/jira/browse/UID2-1894)

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
jon8787: I thought it'd be at a higher level, like keysets. Is is possible to only mock 1 method `reader.getSnapshot()` where  
reader is:    `private final ScopedStoreReader<Map<Integer, Site>> reader;`

Worth doing if it's easy, otherwise this is fine

sunnywu: as discussed offline - will leave it as is

mcollins-ttd (approved):

PR #204, opened 2023-09-11
make refresh times to Core configurable

jon8787:
I want this to make testing alerts easier, but could be useful in other contexts.
(Eventually these configs will come from Core)


mcollins-ttd (approved):

atarassov-ttd (commented):
LGTM

PR #209, opened 2023-09-12
Ian UI d2 1394 update attest failure reaction

Ian-Nara:
[UID2-1395](https://atlassian.thetradedesk.com/jira/browse/UID2-1395)
[UID2-1394](https://atlassian.thetradedesk.com/jira/browse/UID2-1394)


zaiweidu (approved):
LGTM

src/main/java/com/uid2/operator/Main.java
```diff
@@ -96,9 +96,9 @@ public Main(Vertx vertx, JsonObject config) throws Exception {
         if (coreAttestUrl != null) {
             String coreApiToken = this.config.getString(Const.Config.CoreApiTokenProp);
             Duration disableWaitTime = Duration.ofHours(this.config.getInteger(Const.Config.FailureShutdownWaitHoursProp, 120));
```
thomasm-ttd, line 98: This period is probably too long - 120 hours = 5 days. Part of ticket 1395 is to work with ProdSec to get this time, but it should not be configurable - we don't want someone being able to modify it if they are trying to compromise the operator

src/main/java/com/uid2/operator/vertx/OperatorShutdownHandler.java
```diff
@@ -0,0 +1,58 @@
...
+        if (response.right().equals("Unauthorized")) {
+            LOGGER.error("core attestation failed due to invalid operator key. shutting down operator");
+            System.exit(1);
+        } else {
+            try {
+                JsonObject respJson = new JsonObject(response.right());
+                if (respJson.equals(AuthMiddleware.UnauthorizedResponse)) {
+                    LOGGER.error("core attestation failed due to invalid operator key. shutting down operator");
+                    System.exit(1);
+                }else if(respJson.equals(new JsonObject().put("status", "The enclave identifier is unknown"))) {
```
thomasm-ttd, line 38: Has core been updated to return that message? I think the message should be more generic, such as Attestation Failed. There are other reasons other than the enclave id is not known

Ian-Nara: Core will return either "attestation failure", "The payload is ill-formatted", "Cannot verify payload with the signature", "Cannot verify the certificate chain", "The enclave identifier is unknown", or "Unknown reason"

src/main/java/com/uid2/operator/Main.java
```diff
@@ -245,8 +246,6 @@ private ICloudStorage wrapCloudStorageForOptOut(ICloudStorage cloudStorage) {
     private void run() throws Exception {
         Supplier<Verticle> operatorVerticleSupplier = () -> {
             UIDOperatorVerticle verticle = new UIDOperatorVerticle(config, this.clientSideTokenGenerate, siteProvider, clientKeyProvider, clientSideKeypairProvider, getKeyManager(), saltProvider, optOutStore, Clock.systemUTC(), _statsCollectorQueue, new SecureLinkValidatorService(this.serviceLinkProvider));
-            if (this.disableHandler != null)
```
thomasm-ttd, line 248: Do we still use the disableHander in the vertx? If it doesn't get set, it should be removed.

thomasm-ttd (commented):
One small comment

src/main/java/com/uid2/operator/Main.java
Ian-Nara: removed this

src/main/java/com/uid2/operator/Main.java
Ian-Nara: reached out to prodsec, will get a security review setup

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -1823,24 +1844,54 @@ void disableOnDeauthorization(Vertx vertx, VertxTestContext testContext) {
             assertEquals(200, response.statusCode());
 
             // Revoke auth
-            this.operatorDisableHandler.handleResponseStatus(401);
+            try {
```
thomasm-ttd, line 1847: This test and the ones below are really testing the behaviour of the OperatorShutdownHandler. They should be moved to a different file that unit tests the OperatoprShutdownHandler only.
For the tests in UIDOperatorVerticleTest, I am ok with removing the tests that test the old disableHandler

Ian-Nara: Fixed

thomasm-ttd (approved):

PR #214, opened 2023-09-14
Push image to GCP Registry

lunwang-ttd:
## Changes
* Push Docker image to both GitHub Registry and GCP Registry, they use the same image digest. It means when create a new enclave ID, you can still go to the GitHub packages page to find the digest.
* Currently it uses test project, should switch to production once it's ready

### DIFF
--- a/.github/workflows/publish-gcp-oidc-enclave-docker.yaml
+++ b/.github/workflows/publish-gcp-oidc-enclave-docker.yaml
@@ -145,18 +145,12 @@ jobs:
             JAR_VERSION=${{ steps.version.outputs.new_version }}
             IMAGE_VERSION=${{ steps.version.outputs.new_version }}
 
-      - name: Auth GCP
-        id: auth-gcp
-        uses: 'google-github-actions/auth@v1'
-        with:
-          credentials_json: '${{ secrets.GCP_REGISTRY_CREDENTIAL }}'
-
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-gcp-oidc-enclave-docker.yaml
```diff
@@ -84,6 +86,13 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: Log in to the GCP Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ${{ env.GCP_REGISTRY }}
+          username: _json_key
+          password: ${{ secrets.GCP_REGISTRY_CREDENTIAL }}
```
yishi-ttd, line 94: What is this GCP_REGISTRY_CREDENTIAL? How we produce it and save it in secret?

Can we leverage the exported service account json?

@abuabraham-ttd mentioned he may want to only maintain one service account for github action in future.

lunwang-ttd: It's similar to the github-ci service-account, it's also a service account key (JSON).

Sure, we can easily change it after the GCP setup is done.

yishi-ttd (approved):

PR #221, opened 2023-09-21
refine ssl template

yishi-ttd:
1. Put `ssl` `private_key` `certificate` to input optional variable, if specified, we will create cert and enable ssl on LB.
2. Do not use http_redirect, use http_forward instead (by default true).


lunwang-ttd (approved):
LGTM

PR #227, opened 2023-09-22
Client plaintext key removal

gmsdelmundo:


src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -156,9 +155,8 @@ protected void fakeAuth(int siteId, Role... roles) {
     }
     protected void fakeAuth(int siteId, Instant created, Role... roles) {
         ClientKey clientKey = new ClientKey(
-                "test-key",
-                "UID2-C-L-999-fCXrMM.fsR3mDqAXELtWWMS+xG1s7RdgRTMqdOH2qaAo=",
```
vishalegbert-ttd, line 160: NIT maybe the plaintext key should be in a constant and we should use key hasher to generate the hash and salt? Future readers might be confused as to how the plaintext key and these values are related

vishalegbert-ttd (approved):

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
gmsdelmundo: Oh good point, let me update it

PR #230, opened 2023-09-26
Replace "api_token" wording with "operator_key" in doc

yishi-ttd:
Note: we will not update env input `API_TOKEN_SECRET_NAME` for now.


lunwang-ttd (approved):
LGTM

PR #242, opened 2023-10-09
Updating shared in operator

cody-constine-ttd:


Ian-Nara (approved):
LGTM and tests passed locally

PR #243, opened 2023-10-10
Azure CC attestation provider

lunwang-ttd:
Add Azure CC (Confidential Container) attestation provider, major changes are as follows:

- Create Azure CC UID2 core client
- Build Azure CC Docker image
- Update e2e scripts to support local testing

Verified with actual Azure Confidential Container and local stack environment


e2e/docker/localstack/s3/core/operators/operators.json
```diff
@@ -62,6 +62,21 @@
...
+    "key": "UID2-azure-cc-operator-key",
+    "name": "Publisher (Private, Azure CC)",
+    "contact": "Publisher (Private, Azure CC)",
+    "protocol": "azure-cc",
+    "created": 1618873215,
+    "disabled": false,
+    "roles": [
+      "OPERATOR"
+    ],
+    "site_id": 124,
```
yishi-ttd, line 75: nit: use a different site_id with gcp.

scripts/azure-cc/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+RUN tar xzvf /app/static.tar.gz --no-same-owner --no-same-permissions && rm -f /app/static.tar.gz
+
+COPY ./entrypoint.sh /app/
+RUN chmod a+x /app/entrypoint.sh
+
+RUN adduser -D uid2-operator && mkdir -p /opt/uid2 && chmod 777 -R /opt/uid2 && mkdir -p /app && chmod 705 -R /app && mkdir -p /app/file-uploads && chmod 777 -R /app/file-uploads
+USER uid2-operator
+
+CMD ["/app/entrypoint.sh"]
```
yishi-ttd, line 31: nit: no EOL in file

scripts/azure-cc/conf/logback.xml
```diff
@@ -0,0 +1,15 @@
+<configuration>
```
yishi-ttd, line 1: Can we reference the same logback conf file with GCP to avoid duplication?

e2e/e2e.sh
```diff
@@ -2,11 +2,19 @@
...
+CORE_VERSION=2.14.5-SNAPSHOT-default
 OPTOUT_VERSION=2.6.18-60727cf243-default
 
+# GCP OIDC enclave configs
+TEST_GCP_OIDC=false
+IMAGE_HASH=
+
+# Azure CC enclave configs
+TEST_AZURE_CC=false
+AZURE_CC_POLICY_DIGEST=
```
yishi-ttd, line 16: Add a todo here. Eventually we should be able to derive DIGEST via IMAGE_HASH input.

e2e/start_azure_cc_enclave.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+set -ex
+
+echo "Please start Azure CC manually, automation is not implemented yet"
```
yishi-ttd, line 4: nit: no EOL, same below

e2e/docker/localstack/s3/core/operators/operators.json
lunwang-ttd: I think it should be better to also use 124, because site_id 124 is defined for Publisher and it doesn't have harm to use the same site id by multiple operators

e2e/e2e.sh
lunwang-ttd: But the deployment template also matters for the digest besides IMAGE_HASH.
Do you mean we can also derive the corresponding template by IMAGE_HASH?

e2e/start_azure_cc_enclave.sh
lunwang-ttd: updated

scripts/azure-cc/conf/logback.xml
yishi-ttd: OK. Let's leave it as is, but open a ticket to consolidate those log config files later. Andrei also suggests us to use shared conf if possible.

e2e/e2e.sh
lunwang-ttd: Got it, but the input to e2e script could be template file and digest which are generated by CI pipeline.
Let me add a todo, we can decide the input later

scripts/azure-cc/conf/logback.xml
lunwang-ttd: Sure, UID2-2066 is created

e2e/e2e.sh
lunwang-ttd: Updated

yishi-ttd (approved):

PR #246, opened 2023-10-13
UID2-997 CSTG metric based monitoring

jon8787:


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1577,12 +1580,20 @@ private RefreshResponse refreshIdentity(RoutingContext rc, String tokenStr) {
         return this.idService.refreshIdentity(refreshToken);
     }
 
+    public static String getSiteName(ISiteStore siteStore, int siteId) {
+        if (siteStore == null) return "null"; //this is expected if CSTG is not enabled, eg for private operators
```
sunnywu, line 1584: is "null" a convention or calling it "unknown" better than null?

jon8787: saw "unknown" used elsewhere so I've changed it to that

mcollins-ttd (approved):

PR #248, opened 2023-10-18
Return opted out token on CSTG refresh

mcollins-ttd:
If a user has opted out after token generation, we should return an opted out token upon refresh.

### DIFF
--- a/src/main/java/com/uid2/operator/service/UIDOperatorService.java
+++ b/src/main/java/com/uid2/operator/service/UIDOperatorService.java
@@ -21,6 +21,9 @@ import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.util.*;
 
+import static com.uid2.operator.IdentityConst.ClientSideTokenGenerateOptOutIdentityForEmail;
+import static com.uid2.operator.IdentityConst.ClientSideTokenGenerateOptOutIdentityForPhone;
+
 public class UIDOperatorService implements IUIDOperatorService {
     public static final String IDENTITY_TOKEN_EXPIRES_AFTER_SECONDS = "identity_token_expires_after_seconds";
     public static final String REFRESH_TOKEN_EXPIRES_AFTER_SECONDS = "refresh_token_expires_after_seconds";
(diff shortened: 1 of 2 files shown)


src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -274,40 +274,37 @@ private void sendTokenGenerate(String apiVersion, Vertx vertx, String v1GetParam
         }
     }
 
-    private void sendTokenRefresh(String apiVersion, Vertx vertx, String refreshToken, String v2RefreshDecryptSecret, int expectedHttpCode,
+    private void sendTokenRefresh(String apiVersion, Vertx vertx, VertxTestContext testContext, String refreshToken, String v2RefreshDecryptSecret, int expectedHttpCode,
```
sunnywu, line 277: is the whole point of parsing testContext here is so it would automatically only proceed to assert only when the refresh api calls succeeds and otherwise it would assert fail/break test automatically?

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -129,6 +150,20 @@ public RefreshResponse refreshIdentity(RefreshToken token) {
...
+            case Email:
+                return InputUtil.InputVal.validEmail(
+                        ClientSideTokenGenerateOptOutIdentityForEmail,
+                        ClientSideTokenGenerateOptOutIdentityForEmail);
+            case Phone:
+                return InputUtil.InputVal.validPhone(
+                        ClientSideTokenGenerateOptOutIdentityForPhone,
+                        ClientSideTokenGenerateOptOutIdentityForPhone);
+        }
+        throw new IllegalArgumentException("Invalid identity type " + identityType);
```
sunnywu, line 164: is it serious enough we should throw an exception? if identityType is unknown maybe it's best to just assume email instead of raising exception in this case?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -274,40 +274,37 @@ private void sendTokenGenerate(String apiVersion, Vertx vertx, String v1GetParam
...
-                            handler.handle(tryParseResponse(ar.result()));
+                            handler.handle(tryParseResponse(response));
                         }
-                    });
+                    })));
         } else {
-            get(vertx, "v1/token/refresh?refresh_token=" + urlEncode(refreshToken), ar -> {
-                assertTrue(ar.succeeded());
-                HttpResponse<Buffer> response = ar.result();
+            get(vertx, "v1/token/refresh?refresh_token=" + urlEncode(refreshToken), testContext.succeeding(response -> testContext.verify(() -> {
```
jon8787, line 299: should we add Jingyi's changes to our get() method in Operator?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
jon8787: I think you need this if there are any async calls at all. Otherwise the test completes and passes without waiting for the async code to complete.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -3340,17 +3357,75 @@ else if(identityType == IdentityType.Phone) {
...
+
+                () -> assertEquals(siteId, advertisingToken.publisherIdentity.siteId, "Advertising token site ID is incorrect"),
+                () -> assertEquals(siteId, refreshToken.publisherIdentity.siteId, "Refresh token site ID is incorrect"),
+
+                () -> assertArrayEquals(advertisingId, advertisingToken.userIdentity.id, "Advertising token ID is incorrect"),
+                () -> assertArrayEquals(firstLevelHash, refreshToken.userIdentity.id, "Refresh token ID is incorrect")
+        );
+    }
+
+    private void assertAreClientSideGeneratedOptOutTokens(AdvertisingToken advertisingToken, RefreshToken refreshToken, int siteId, IdentityType identityType) {
```
jon8787, line 3391: can we reuse any code for these 2 methods?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: The problem I was facing was that due to the lack of `testContext.succeeding`, each time as assert failed in the callback the test would keep running and time out after 30s. With `succeeding`, the test fails immediately.

src/main/java/com/uid2/operator/service/UIDOperatorService.java
mcollins-ttd: If we're going to do that I'd like to add an `assert` as well. That way the assert will fail for any tests that call this method with the new enum value, but we can continue to return email in production. @jon8787 ?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: That would be a big change, maybe for another PR.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: It would be possible but I'm not sure it's worth it, might end up making the methods harder to follow.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
jon8787: even if it's just a new overload for now?

src/main/java/com/uid2/operator/service/UIDOperatorService.java
mcollins-ttd: Updated to return a fallback value and assert

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: Extracted a method from these two

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: As discussed offline, we will look at making this change in another PR. This change was just so I could iterate without having to wait 30 seconds each time.

jon8787 (approved):

sunnywu (approved):

PR #249, opened 2023-10-18
UID2-2094 Add token response metric recording when cstg endpoint processing fails some checks

sunnywu:
…ls some checks

### DIFF
--- a/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
+++ b/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
@@ -308,6 +308,12 @@ public class UIDOperatorVerticle extends AbstractVerticle {
         }
     }
 
+    private void SendErrorResponseAndRecordStats(String errorStatus, int statusCode, RoutingContext rc, String message, Integer siteId, TokenResponseStatsCollector.Endpoint endpoint, TokenResponseStatsCollector.ResponseStatus responseStatus)
+    {
+        ResponseUtil.Error(errorStatus, statusCode, rc, message);
+        recordTokenResponseStats(siteId, endpoint, responseStatus);
+    }
+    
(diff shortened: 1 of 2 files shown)


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -317,12 +318,13 @@ private void handleClientSideTokenGenerateImpl(RoutingContext rc) throws NoSuchA
             body = rc.body().asJsonObject();
         } catch (DecodeException ex) {
             ResponseUtil.Error(ResponseStatus.ClientError, 400, rc, "json payload is not valid");
+            recordTokenResponseStats(null, TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.BadJsonPayload);
```
jon8787, line 321: can we make a method that calls both ResponseUtil.Error and recordTokenResponseStats? Seems like it's never correct to call the 1st without the 2nd

sunnywu: there are 38 calls to com.uid2.operator.service.ResponseUtil#Error so i can create a new ticket about the proposed improvement but don't really wanna make this change in this small PR? 

There are quite a lot of instance where we don't call recordTokenResponseStats and should confirm the reason before making wholesale changes

jon8787: Right but I'm not saying change 38 calls, just the new ones you've introduced. The 38 calls can be a future thing to investigate.

sunnywu: @jon8787 done please have a look again

jon8787 (approved):

PR #250, opened 2023-10-18
E2E workflow for azure-cc

yishi-ttd:
Add logic to run Azure enclave E2E verification.
Fixed `jq | tee` issue: `tee`  redirection not work as expected. It will introduce random failure.

A successful run example: https://github.com/IABTechLab/uid2-operator/actions/runs/6561082046


lunwang-ttd (approved):
LGTM

PR #253, opened 2023-10-25
fix to private operator throwing for V2 endpoints due to CSTG metrics

jon8787:


src/main/java/com/uid2/operator/Main.java
```diff
@@ -143,6 +142,8 @@ public Main(Vertx vertx, JsonObject config) throws Exception {
             this.serviceLinkProvider = new RotatingServiceLinkStore(fsStores, new GlobalScope(new CloudPath(serviceLinkMdPath)));
         }
 
+        this.siteProvider = clientSideTokenGenerate ? new RotatingSiteStore(fsStores, new GlobalScope(new CloudPath(sitesMdPath))) : null;
```
thomasm-ttd, line 145: siteProvider gets passed to a number of other places, and if it is null, it is going to fail. For example, in line 307 of main.

jon8787: line 307 doesn't get hit if clientSideTokenGenerate is false

vishalegbert-ttd (approved):

PR #254, opened 2023-10-25
Load azure secret from vault in JAVA code

yishi-ttd:
Load operator key from JAVA code, instead of shell script.


src/main/java/com/uid2/operator/store/CloudSyncOptOutStore.java
```diff
@@ -58,7 +58,7 @@ public CloudSyncOptOutStore(Vertx vertx, ICloudStorage fsLocal, JsonObject jsonC
             this.remoteApiPort = -1 == url.getPort() ? 80 : url.getPort();
             this.remoteApiHost = url.getHost();
             this.remoteApiPath = url.getPath();
-            this.remoteApiBearerToken = "Bearer " + jsonConfig.getString(Const.Config.OptOutApiTokenProp);
+            this.remoteApiBearerToken = "Bearer " + operatorKey;
```
thomasm-ttd, line 61: Why are we changing this? It is also used directly above on line 55

src/main/java/com/uid2/operator/operatorkey/ConfigOperatorKeyRetriever.java
```diff
@@ -0,0 +1,17 @@
...
+public class ConfigOperatorKeyRetriever implements IOperatorKeyRetriever {
+    private final JsonObject config;
+
+    public ConfigOperatorKeyRetriever(JsonObject config) {
+        this.config = config;
+    }
+
+    @Override
+    public String retrieve() {
+        return this.config.getString(Const.Config.CoreApiTokenProp);
```
thomasm-ttd, line 15: The name of the property and the class name difference will cause confusion - the OperatorKeyRetriever reads the CoreApiTokenProp? They should both be called OperatorKey or CoreApiToken

src/main/java/com/uid2/operator/store/CloudSyncOptOutStore.java
yishi-ttd: L55 is a different config.
Currently we will set the same operator key for both "OptOutApiToken" and "CoreApiToken".

The value of "OptOutApiToken" should also be fetched from vault - actually it's just operator key.

src/main/java/com/uid2/operator/operatorkey/ConfigOperatorKeyRetriever.java
yishi-ttd: CoreApiTokenProp(core_api_token) is an existing config in shared and I don't want to change this to break existing logic.
E.g. currently AWS/GCP will set  `core_api_token` and `optout_api_token`, they could still work after this change.

src/main/java/com/uid2/operator/store/CloudSyncOptOutStore.java
thomasm-ttd: Ok, but this would mean we are actually getting rid of the config setting for optout_api_token. In that case, we need to remove it from the code, and from all the config. This could be another ticket, but I don't think we should leave unused config settings in the code / config files

src/main/java/com/uid2/operator/Main.java
```diff
@@ -484,4 +489,19 @@ private Map.Entry<UidCoreClient, UidOptOutClient> createUidClients(Vertx vertx,
         UidOptOutClient optOutClient = new UidOptOutClient(clientApiToken, CloudUtils.defaultProxy, enforceHttps, attestationTokenRetriever);
         return new AbstractMap.SimpleEntry<>(coreClient, optOutClient);
     }
+
+    private IOperatorKeyRetriever createOperatorKeyRetriever() throws Exception {
+        var enclavePlatform = this.config.getString("enclave_platform", "");
```
thomasm-ttd, line 494: We need to ensure that we can diagnose an operator that does not have these settings set. This is going to throw an exception and shut the operator down. There must be enough information so that whoever is trying to start the operator can determine that those settings are not set.

.github/workflows/publish-azure-cc-enclave-docker.yaml
```diff
@@ -134,6 +134,7 @@ jobs:
 
       - name: Test with Trivy vulnerability scanner
         uses: aquasecurity/trivy-action@master
+        if: inputs.publish_vulnerabilities == 'true'
```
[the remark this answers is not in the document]
yishi-ttd: First, I totally agree we need to run vulnerable check. The `publish_vulerabilities` check is already there.

But accourding to this workflow logic, this step should not run if `publish_vulnerabilities` is set to false - it depends on previous step's output data. ( See L 119)

src/main/java/com/uid2/operator/store/CloudSyncOptOutStore.java
yishi-ttd: I will remove OptOutApiTokenProp config.
This is the only place that will use it.

yishi-ttd: As for the places to set this env, I will create a ticket to track.

src/main/java/com/uid2/operator/Main.java
yishi-ttd: We will throw IllegalArgument Exception from AzureVaultOperatorKeyRetriever, if `vaultName` or `secretName` is not set: https://github.com/IABTechLab/uid2-attestation-azure/pull/6/files#diff-e825008e6ad458d64e6842bbe6a24e96cedca3ee6cf25f0b4aaaf00635f3e74fR18

There should be enough info for diagnosis.

thomasm-ttd (approved):

PR #259, opened 2023-10-27
Improve response status codes and logging

alex-yau-ttd:
- Introduce `ClientInputException` to help ensure that the correct status codes are returned
- Log more detail when throwing exception
- Replace `IllegalArgumentException` with `ClientInputException` when it is a genuine 400. Some `IllegalArgumentException`s remain, such as ones thrown by `Base64.getDecoder().decode()`
- Refactor SendErrorResponseAndRecordStats by moving to `ResponseUtil.java` and splitting into separate methods for 4xx (warnings) and 500 (error)
- Use `SendClientErrorResponseAndRecordStats` and `SendServerErrorResponseAndRecordStats` for CSTG, generate and refresh endpoints
- Move `ResponseStatus` from `UIDOperatorVerticle` into `ResponseUtil`

Note: I've screwed up the commit history (failed revert, etc.), so let me know if something is not clear.


src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
```diff
@@ -72,16 +73,17 @@ private byte[] encodeV3(AdvertisingToken t, KeysetKey masterKey, KeysetKey siteK
...
     @Override
     public RefreshToken decodeRefreshToken(String s) {
-        final byte[] bytes = EncodingUtils.fromBase64(s);
-        final Buffer b = Buffer.buffer(bytes);
-
-        if (b.getByte(1) == TokenVersion.V3.rawVersion) {
-            return decodeRefreshTokenV3(b, bytes);
-        } else if (b.getByte(0) == TokenVersion.V2.rawVersion) {
-            return decodeRefreshTokenV2(b);
+        if (s != null && !s.isEmpty()) {
```
alex-yau-ttd, line 76: This was caused a `NullPointerException`, hence why the method that calls this (`refreshIdentity()`) had to catch a `Throwable`.  With this change, now `refreshIdentity()` only needs to catch `ClientInputException`.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -877,7 +878,7 @@ private void handleTokenGenerateV2(RoutingContext rc) {
     private void handleTokenGenerate(RoutingContext rc) {
         final InputUtil.InputVal input = this.getTokenInput(rc);
         if (input == null || !input.isValid()) {
-            rc.fail(400);
+            ResponseUtil.Error(ResponseStatus.InvalidToken, 400, rc, "Invalid Token presented " + input);
```
vishalegbert-ttd, line 881: I think this is an invalid email/phone numebr rather than an invalid token? Not sure if we should log it out

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1095,7 +1096,7 @@ private void handleIdentityMap(RoutingContext rc) {
                 final MappedIdentity mappedIdentity = this.idService.map(input.toUserIdentity(this.identityScope, 0, now), now);
                 rc.response().end(EncodingUtils.toBase64String(mappedIdentity.advertisingId));
             } else {
-                rc.fail(400);
+                ResponseUtil.Error(ResponseStatus.InvalidToken, 400, rc, "Invalid Token presented " + input);
```
vishalegbert-ttd, line 1099: I think this is an invalid email/phone numebr rather than an invalid token? Not sure if we should log it out

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -903,7 +904,7 @@ private void handleTokenGenerate(RoutingContext rc) {
     private void handleTokenRefresh(RoutingContext rc) {
         final List<String> tokenList = rc.queryParam("refresh_token");
         if (tokenList == null || tokenList.size() == 0) {
-            rc.fail(400);
+            ResponseUtil.ClientError(rc, "Invalid refresh_token");
```
vishalegbert-ttd, line 907: Maybe a better message is no refresh token given or something along those lines?

src/main/java/com/uid2/operator/vertx/ClientInputException.java
```diff
@@ -0,0 +1,11 @@
+package com.uid2.operator.vertx;
+
+public class ClientInputException extends RuntimeException  {
```
vishalegbert-ttd, line 3: I wonder if a better name is ClientInputValidationException

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -751,6 +751,8 @@ private void handleTokenValidateV1(RoutingContext rc) {
             } else {
                 ResponseUtil.Success(rc, Boolean.FALSE);
             }
+        } catch (ClientInputException cie) {
+            ResponseUtil.Error(ResponseStatus.InvalidToken, 400, rc, "Invalid Token presented");
```
vishalegbert-ttd, line 755: Seems like this will log an error rather than a warning?

alex-yau-ttd: yeah this is the existing pattern for when the token is invalid. `ResponseUtil.ClientError()` also logs an error. Would you rather them all be set to warnings? I see that is being a new method `ResponseUtil.Warning()`

vishalegbert-ttd: I think all client input validation failures should be logged as warning to keep the error logs as clean as possible. I'd prefer to change ClientError to log a warning instead

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -865,8 +867,7 @@ private void handleTokenGenerateV2(RoutingContext rc) {
                     recordTokenResponseStats(clientKey.getSiteId(), TokenResponseStatsCollector.Endpoint.GenerateV2, TokenResponseStatsCollector.ResponseStatus.Success);
                 }
             }
-        } catch (IllegalArgumentException iae) {
-            LOGGER.warn("request body contains invalid argument(s)", iae);
+        } catch (ClientInputException cie) {
             ResponseUtil.ClientError(rc, "request body contains invalid argument(s)");
```
vishalegbert-ttd, line 871: Could you please add this line in this catch block:
recordTokenResponseStats(clientKey.getSiteId(), TokenResponseStatsCollector.Endpoint.GenerateV2, TokenResponseStatsCollector.ResponseStatus.MissingParams);

jon8787: Should be `SendErrorResponseAndRecordStats` no?
We should hardly ever need these as 2 separate calls.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: good catch - this looks to be an email/hash - I don't see where phone number comes into play? The input only comes back null if both email and email hash are supplied.  I can't really tell how `input.isValid()` works. So should I change to
```
ResponseUtil.ClientError(rc, "Required Parameter Missing: exactly one of email or email_hash must be specified");
```

vishalegbert-ttd:
Could you also add 
recordTokenResponseStats(clientKey.getSiteId(), TokenResponseStatsCollector.Endpoint.GenerateV2, TokenResponseStatsCollector.ResponseStatus.InvalidPayload); to V2PayloadHandler.handleTokenGenerate where !request.isValid()

vishalegbert-ttd:
Could you also add 
recordTokenResponseStats(clientKey.getSiteId(), TokenResponseStatsCollector.Endpoint.RefreshV2, TokenResponseStatsCollector.ResponseStatus.InvalidPayload); to V2PayloadHandler.handleTokenRefresh where !request.isValid()

jon8787:
> Could you also add recordTokenResponseStats(clientKey.getSiteId(), TokenResponseStatsCollector.Endpoint.RefreshV2, TokenResponseStatsCollector.ResponseStatus.InvalidPayload); to V2PayloadHandler.handleTokenRefresh where !request.isValid()

`SendErrorResponseAndRecordStats`

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
vishalegbert-ttd: Yeah perhaps, I really don't like that we log client errors at the Error level

alex-yau-ttd: do we need a new `SendWarningResponseAndRecordStats()`?

That would probably be better than a `SendResponseAndRecordStats()` that uses `statusCode` to determine if it should warn (for 400s) or error (for 500s).

What do people think?

jon8787: I'd prefer if the caller doesn't have to think about what level they need to log at, and the method knows what to do.

src/main/java/com/uid2/operator/model/IdentityScope.java
```diff
@@ -12,15 +14,15 @@ public static IdentityScope fromValue(int value) {
         switch (value) {
             case 0: return UID2;
             case 1: return EUID;
-            default: throw new IllegalArgumentException();
+            default: throw new ClientInputException("Invalid value for IdentityScope: " + value);
```
asloobq, line 17: Just curious if there a code flow which doesn't originate from a client side request that can trigger this ? For example, a cron job.  Because in that case it should be upto the caller to decide whether it qualifies for a ClientInputException.

alex-yau-ttd: Good question. Tracking the usages, the only code paths I can see in operator are from client side requests (`UIDOperatorVerticle.java`), benchmarks (`TokenEndecBenchmark.java` ) and tests. I'm not very familiar with cron jobs - how would I check if cron jobs can hit this?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: to clarify, should current 500s like 
```
LOGGER.error("Unknown error while generating token", e);
rc.fail(500);
```
also be changed to use `SendErrorResponseAndRecordStats`? All usages of `SendErrorResponseAndRecordStats` are currently 400s

jon8787: Does `SendErrorResponseAndRecordStats` call `rc.fail` though?

Ideally current 500s should call it as well, assuming we keep these as 500s and don't get an unexpected change in behavior.

alex-yau-ttd: > Does SendErrorResponseAndRecordStats call rc.fail though?
It doesn't now, but it could. Otherwise could just do 
```
SendErrorResponseAndRecordStats(...);
rc.fail(500);
```

jon8787: Sounds good if `SendErrorResponseAndRecordStats` does everything that's needed, including rc.fail

alex-yau-ttd:
> > Could you also add recordTokenResponseStats(clientKey.getSiteId(), TokenResponseStatsCollector.Endpoint.RefreshV2, TokenResponseStatsCollector.ResponseStatus.InvalidPayload); to V2PayloadHandler.handleTokenRefresh where !request.isValid()
> 
> `SendErrorResponseAndRecordStats`

`SendErrorResponseAndRecordStats` is currently a private method in `UIDOperatorVerticle.java` - I assume this method, along with `recordTokenResponseStats` should go in a separate class, given we want to use it in `V2PayloadHandler` which `UIDOperatorVerticle` already has one of. Would it make sense to go in `TokenResponseStatsCollector.java` as static methods? Or perhaps in `V2PayloadHandler` itself? 

I think `TokenResponseStatsCollector` makes the most sense
@jon8787 @vishalegbert-ttd

src/main/java/com/uid2/operator/vertx/ClientInputException.java
alex-yau-ttd: done

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: fixed

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: I've changed `SendErrorResponseAndRecordStats` to handle the logic, depending on the status code. I've changed all instances of the previous pattern for generate, refresh and CSTG handlers.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: changed the message to be the same as the v1 endpoint

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: I've inverted the `if` in this method and split into multiple conditionals, in order to give more detailed logging. Let me know if that looks right!

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: I've split this into multiple conditionals, in order to give more detailed logging. Let me know if that looks right!

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: sorry, missed it. fixing now

alex-yau-ttd: done. I see there are also some 401s - should these warn instead of error as well?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -292,9 +292,7 @@ private void handleClientSideTokenGenerate(RoutingContext rc) {
         try {
             handleClientSideTokenGenerateImpl(rc);
         } catch (Exception e) {
-            LOGGER.error("Unknown error while handling client side token generate", e);
-            recordTokenResponseStats(null, TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.Unknown);
-            rc.fail(500);
+            SendErrorResponseAndRecordStats(ResponseStatus.UnknownError, 500, rc, "Unknown error while handling client side token generate", null, TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.Unknown);
```
vishalegbert-ttd, line 295: Are we no longer logging the exception here?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -310,7 +308,12 @@ private Set<String> getDomainNameListForClientSideTokenGenerate(ClientSideKeypai
 
     private void SendErrorResponseAndRecordStats(String errorStatus, int statusCode, RoutingContext rc, String message, Integer siteId, TokenResponseStatsCollector.Endpoint endpoint, TokenResponseStatsCollector.ResponseStatus responseStatus)
     {
-        ResponseUtil.Error(errorStatus, statusCode, rc, message);
+        if (statusCode == 400 || statusCode == 403) {
```
vishalegbert-ttd, line 311: I think 400 >= && <=499 is safer, same for 500

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -396,13 +395,11 @@ private void handleClientSideTokenGenerateImpl(RoutingContext rc) throws NoSuchA
...
             if (ivBytes.length != 12) {
-                ResponseUtil.Error(ResponseStatus.ClientError, 400, rc, "bad iv");
-                recordTokenResponseStats(clientSideKeypair.getSiteId(), TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.BadIV);
+                SendErrorResponseAndRecordStats(ResponseStatus.ClientError, 400, rc, "bad iv", clientSideKeypair.getSiteId(), TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.BadIV);
                 return;
             }
         } catch (IllegalArgumentException e) {
-            ResponseUtil.Error(ResponseStatus.ClientError, 400, rc, "bad iv");
-            recordTokenResponseStats(clientSideKeypair.getSiteId(), TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.BadIV);
+            SendErrorResponseAndRecordStats(ResponseStatus.ClientError, 400, rc, "bad iv", clientSideKeypair.getSiteId(), TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.BadIV);
```
vishalegbert-ttd, line 402: Exception logging?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -415,17 +412,15 @@ private void handleClientSideTokenGenerateImpl(RoutingContext rc) throws NoSuchA
             System.arraycopy(encryptedPayloadBytes, 0, ivAndCiphertext, 12, encryptedPayloadBytes.length);
             requestPayloadBytes = decrypt(ivAndCiphertext, 0, sharedSecret, aad);
         } catch (Exception e) {
-            ResponseUtil.Error(ResponseStatus.ClientError, 400, rc, "payload decryption failed");
-            recordTokenResponseStats(clientSideKeypair.getSiteId(), TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.BadPayload);
+            SendErrorResponseAndRecordStats(ResponseStatus.ClientError, 400, rc, "payload decryption failed", clientSideKeypair.getSiteId(), TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.BadPayload);
```
vishalegbert-ttd, line 415: Exception logging?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -415,17 +412,15 @@ private void handleClientSideTokenGenerateImpl(RoutingContext rc) throws NoSuchA
...
             return;
         }
 
         final JsonObject requestPayload;
         try {
             requestPayload = new JsonObject(Buffer.buffer(Unpooled.wrappedBuffer(requestPayloadBytes)));
         } catch (DecodeException e) {
-            ResponseUtil.Error(ResponseStatus.ClientError, 400, rc, "encrypted payload contains invalid json");
-            recordTokenResponseStats(clientSideKeypair.getSiteId(), TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.BadPayload);
+            SendErrorResponseAndRecordStats(ResponseStatus.ClientError, 400, rc, "encrypted payload contains invalid json", clientSideKeypair.getSiteId(), TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.BadPayload);
```
vishalegbert-ttd, line 423: Exception logging?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: yeah good catch, it looks like this is one of the only ones that was actually logging the exception. Do you suggest to modify `SendErrorResponseAndRecordStats` to optionally accept an exception that is logged?

jon8787: 500 should always log the exception no? Yeah we should pass it in if that makes sense

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -668,24 +661,24 @@ private void handleTokenRefreshV1(RoutingContext rc) {
...
             final RefreshResponse r = this.refreshIdentity(rc, refreshToken);
-            Integer siteId = rc.get(Const.RoutingContextData.SiteId);
+            siteId = rc.get(Const.RoutingContextData.SiteId);
             if (!r.isRefreshed()) {
                 if (r.isOptOut() || r.isDeprecated()) {
                     ResponseUtil.SuccessNoBody(ResponseStatus.OptOut, rc);
                 } else if (!AuthMiddleware.isAuthenticated(rc)) {
                     // unauthenticated clients get a generic error
-                    ResponseUtil.Error(ResponseStatus.GenericError, 400, rc, "Error refreshing token");
+                    ResponseUtil.Warning(ResponseStatus.GenericError, 400, rc, "Error refreshing token");
```
vishalegbert-ttd, line 677: Should we use SendErrorResponseAndRecordStats here as well?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -751,6 +743,8 @@ private void handleTokenValidateV1(RoutingContext rc) {
             } else {
                 ResponseUtil.Success(rc, Boolean.FALSE);
             }
+        } catch (ClientInputValidationException cie) {
+            ResponseUtil.Warning(ResponseStatus.InvalidToken, 400, rc, "Invalid Token presented");
```
vishalegbert-ttd, line 747: Exception logging?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -725,8 +718,7 @@ private void handleTokenRefreshV2(RoutingContext rc) {
             }
             TokenResponseStatsCollector.recordRefresh(siteProvider, siteId, TokenResponseStatsCollector.Endpoint.RefreshV2, r);
         } catch (Exception e) {
-            LOGGER.error("Unknown error while refreshing token v2", e);
-            ResponseUtil.Error(ResponseStatus.UnknownError, 500, rc, "Service Error");
+            SendErrorResponseAndRecordStats(ResponseStatus.UnknownError, 500, rc, "Unknown error while refreshing token v2", siteId, TokenResponseStatsCollector.Endpoint.RefreshV2, TokenResponseStatsCollector.ResponseStatus.Unknown);
```
vishalegbert-ttd, line 721: Exception logging?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -789,49 +783,47 @@ private void handleTokenValidateV2(RoutingContext rc) {
...
                 //Integer.parseInt(rc.queryParam("privacy_bits").get(0))));
 
                 ResponseUtil.Success(rc, toJsonV1(t));
-                recordTokenResponseStats(clientKey.getSiteId(), TokenResponseStatsCollector.Endpoint.GenerateV1, TokenResponseStatsCollector.ResponseStatus.Success);
+                recordTokenResponseStats(siteId, TokenResponseStatsCollector.Endpoint.GenerateV1, TokenResponseStatsCollector.ResponseStatus.Success);
             }
         } catch (Exception e) {
-            LOGGER.error("Unknown error while generating token v1", e);
-            rc.fail(500);
+            SendErrorResponseAndRecordStats(ResponseStatus.UnknownError, 500, rc, "Unknown error while generating token v1", siteId, TokenResponseStatsCollector.Endpoint.GenerateV1, TokenResponseStatsCollector.ResponseStatus.Unknown);
```
vishalegbert-ttd, line 804: Exception logging?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -847,63 +839,66 @@ private void handleTokenGenerateV2(RoutingContext rc) {
...
                     ResponseUtil.SuccessV2(rc, toJsonV1(t));
-                    recordTokenResponseStats(clientKey.getSiteId(), TokenResponseStatsCollector.Endpoint.GenerateV2, TokenResponseStatsCollector.ResponseStatus.Success);
+                    recordTokenResponseStats(siteId, TokenResponseStatsCollector.Endpoint.GenerateV2, TokenResponseStatsCollector.ResponseStatus.Success);
                 }
             }
-        } catch (IllegalArgumentException iae) {
-            LOGGER.warn("request body contains invalid argument(s)", iae);
-            ResponseUtil.ClientError(rc, "request body contains invalid argument(s)");
+        } catch (ClientInputValidationException cie) {
+            SendErrorResponseAndRecordStats(ResponseStatus.ClientError, 400, rc, "request body contains invalid argument(s)", siteId, TokenResponseStatsCollector.Endpoint.GenerateV2, TokenResponseStatsCollector.ResponseStatus.MissingParams);
```
vishalegbert-ttd, line 861: Exception logging?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -847,63 +839,66 @@ private void handleTokenGenerateV2(RoutingContext rc) {
...
         } catch (Exception e) {
-            LOGGER.error("Unknown error while generating token v2", e);
-            rc.fail(500);
+            SendErrorResponseAndRecordStats(ResponseStatus.UnknownError, 500, rc, "Unknown error while generating token v2", siteId, TokenResponseStatsCollector.Endpoint.GenerateV2, TokenResponseStatsCollector.ResponseStatus.MissingParams);
```
vishalegbert-ttd, line 863: Exception logging?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -847,63 +839,66 @@ private void handleTokenGenerateV2(RoutingContext rc) {
...
             //Integer.parseInt(rc.queryParam("privacy_bits").get(0))));
 
-            recordTokenResponseStats(clientKey.getSiteId(), TokenResponseStatsCollector.Endpoint.GenerateV0, TokenResponseStatsCollector.ResponseStatus.Success);
+            recordTokenResponseStats(siteId, TokenResponseStatsCollector.Endpoint.GenerateV0, TokenResponseStatsCollector.ResponseStatus.Success);
             sendJsonResponse(rc, toJson(t));
 
         } catch (Exception e) {
-            LOGGER.error("Unknown error while generating token", e);
-            rc.fail(500);
+            SendErrorResponseAndRecordStats(ResponseStatus.UnknownError, 500, rc, "Unknown error while generating token", siteId, TokenResponseStatsCollector.Endpoint.GenerateV0, TokenResponseStatsCollector.ResponseStatus.Unknown);
```
vishalegbert-ttd, line 893: Exception logging?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -912,14 +907,13 @@ private void handleTokenRefresh(RoutingContext rc) {
...
-            Integer siteId = rc.get(Const.RoutingContextData.SiteId);
+            siteId = rc.get(Const.RoutingContextData.SiteId);
             if (r.isRefreshed()) {
                 this.recordRefreshDurationStats(siteId, getApiContact(rc), r.getDurationSinceLastRefresh(), rc.request().headers().contains("Origin"));
             }
             TokenResponseStatsCollector.recordRefresh(siteProvider, siteId, TokenResponseStatsCollector.Endpoint.RefreshV0, r);
         } catch (Exception e) {
-            LOGGER.error("Unknown error while refreshing token", e);
-            rc.fail(500);
+            SendErrorResponseAndRecordStats(ResponseStatus.UnknownError, 500, rc, "Unknown error while refreshing token", siteId, TokenResponseStatsCollector.Endpoint.RefreshV0, TokenResponseStatsCollector.ResponseStatus.Unknown);
```
vishalegbert-ttd, line 916: Exception logging?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
vishalegbert-ttd: Actually nah it's fine. The combination of this and TokenResponseStatsCollector.recordRefresh would do the same thing I think

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: I've changed the 401 to a warning, given the changes to `SendErrorResponseAndRecordStats` will warn for any `4xx`

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done - I've logged out exceptions where possible

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done, see `ResponseUtil.java`

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
alex-yau-ttd: done

src/main/java/com/uid2/operator/service/ResponseUtil.java
```diff
@@ -57,13 +59,39 @@ public static void SuccessV2(RoutingContext rc, Object body) {
...
+        if (statusCode >= 400 && statusCode <= 499) {
+            Warning(errorStatus, statusCode, rc, message);
+        } else if (statusCode >= 500 && statusCode <= 599) {
+            Error(errorStatus, statusCode, rc, message);
+            rc.fail(statusCode);
+        }
+        recordTokenResponseStats(siteId, endpoint, responseStatus, siteProvider);
+    }
+
+    public static void SendErrorResponseAndRecordStats(String errorStatus, int statusCode, RoutingContext rc, String message, Integer siteId, TokenResponseStatsCollector.Endpoint endpoint, TokenResponseStatsCollector.ResponseStatus responseStatus, ISiteStore siteProvider, Exception exception)
```
vishalegbert-ttd, line 76: Duplicating this doesn't look great, I think it's because we're trying to make this function report on both Warning and Error

I know @jon8787 suggested this so that the caller doesn't have to think about it, but the caller is already thinking about the response status, status code, whether or not it has a message etc. I think it should be easy enough for the caller to choose between Warning or Error

To help clean this up, what do you think of the following suggestion?

- SendClientErrorResponseAndRecordStats(errorStatus, statusCode, rc, message, siteId, endpoint, responseStatus, siteProvider)
- SendErrorResopnseAndRecordStats(rc, message, siteId, endpoint, responseStatus, siteProvider, exception)

SendClientErrorResponseAndRecordStats doesn't need an exception, usually client errors should be fully known. The message of the client error should be descriptive enough to easily track down

SendErrorResopnseAndRecordStats should always be unknown error, perhaps the message can briefly describe what might have gone wrong, there should be some associated exception

I think this will make the code cleaner

alex-yau-ttd: yeah I think given that client errors and server error tend towards different parameters and require method overloading, I think having separate methods with less parameters will make the code easier to understand. I'll action that now.

jon8787: Agree with having different methods if the arguments are different, but those names are confusing. "Error" sounds like a superset of "ClientError". maybe ServerError or InternalError etc?

alex-yau-ttd: > SendErrorResopnseAndRecordStats(rc, message, siteId, endpoint, responseStatus, siteProvider, exception)
Is the logic behind dropping errorStatus and statusCode the following?
* All `errorStatus` should be unknown
* All statusCode should be 500? Should this method no longer cater for any 5xx?

vishalegbert-ttd: > Should this method no longer cater for any 5xx?
For now I don't think we return any thing besides 500 so I don't think it's necessary

alex-yau-ttd: done!

alex-yau-ttd: Note that not all uses for `Error` actually have an exception, so I couldn't remove all the overloading

vishalegbert-ttd (approved):

PR #261, opened 2023-10-29
[UID2-2148] Add metrics and logs for Service and Service Links

asloobq:


cody-constine-ttd (approved):

src/test/java/com/uid2/operator/ServiceStoreTest.java
```diff
@@ -33,7 +33,7 @@ public void loadFromEmbeddedResourceStorage() throws Exception {
         assertEquals("testName2", service.getName());
 
         Service s1 = new Service(1, 123, "testName1", Set.of(Role.GENERATOR));
-        Service s2 = new Service(2, 123, "testName2", Set.of(Role.ID_READER));
```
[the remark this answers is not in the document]
asloobq: I changed the [test data](**url**) of the service this unit test depends upon. I changed it to setup a client key locally that can be used for testing service and service _links.

src/main/java/com/uid2/operator/service/ResponseUtil.java
```diff
@@ -81,15 +81,23 @@ public static void Error(String errorStatus, int statusCode, RoutingContext rc,
 
 
     private static void logError(String errorStatus, int statusCode, String message, RoutingContextReader contextReader, String clientAddress) {
-        String errorMessage = "Error response to http request. " + JsonObject.of(
+        JsonObject errorJsonObj = JsonObject.of(
                 "errorStatus", errorStatus,
                 "contact", contextReader.getContact(),
                 "siteId", contextReader.getSiteId(),
                 "path", contextReader.getPath(),
```
thomasm-ttd, line 88: Do not log the path as this may contain sensitive information. Yes we have filters for that, but best not to log it at all.

src/main/java/com/uid2/operator/service/ResponseUtil.java
```diff
@@ -81,15 +81,23 @@ public static void Error(String errorStatus, int statusCode, RoutingContext rc,
 
 
     private static void logError(String errorStatus, int statusCode, String message, RoutingContextReader contextReader, String clientAddress) {
-        String errorMessage = "Error response to http request. " + JsonObject.of(
+        JsonObject errorJsonObj = JsonObject.of(
```
thomasm-ttd, line 84: Can contextReader ever be null?

src/main/java/com/uid2/operator/service/SecureLinkValidatorService.java
```diff
@@ -12,6 +12,8 @@
 
 public class SecureLinkValidatorService {
     public static final String LINK_ID = "link_id";
+    public static final String LINK_NAME = "link_name";
```
thomasm-ttd, line 15: Can we call this service_link_name?

src/main/java/com/uid2/operator/service/SecureLinkValidatorService.java
```diff
@@ -25,6 +27,8 @@ public boolean validateRequest(RoutingContext rc, JsonObject requestJsonObject)
         if (profile instanceof ClientKey) {
             ClientKey clientKey = (ClientKey) profile;
             if (clientKey.getServiceId() != 0) {
+                // For now using the client key, in the future when we have multiple keys, this can be changed to service name
+                rc.put(SERVICE_NAME, clientKey.getName());
```
thomasm-ttd, line 31: I am uncomfortable putting the client key name into the service name. How is this being used? Why not put it into a client key field?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1566,6 +1567,44 @@ private void recordIdentityMapStats(RoutingContext rc, int inputCount, int inval
...
+    private void recordIdentityMapStatsForServiceLinks(RoutingContext rc, String apiContact, int inputCount,
+                                                       int invalidCount, int optoutCount) {
+        // If request is from a service, break it down further by link_id
+        String serviceLinkName = rc.get(SecureLinkValidatorService.LINK_NAME, "");
+        if (!serviceLinkName.isBlank()) {
+            // serviceName will be non-empty as it will be inserted during validation
+            String serviceName = rc.get(SecureLinkValidatorService.SERVICE_NAME);
+            DistributionSummary ds = _identityMapMetricSummaries.computeIfAbsent(serviceName + serviceLinkName,
+                    k -> DistributionSummary.builder("uid2.operator.identity.map.services.inputs")
+                .description("number of emails or email hashes passed to identity map batch endpoint")
```
thomasm-ttd, line 1582: What about phone numbers or phone hashes?

src/main/java/com/uid2/operator/service/ResponseUtil.java
asloobq: right now this method is being called only by the public `Error` method which instantiates the contextReader. So it won't be null

src/main/java/com/uid2/operator/service/SecureLinkValidatorService.java
asloobq: Right now there is only 1 service - AWS Venice and both the service name and client key name are same. 
I could lookup the Service Name from the `ServiceStore` but didn't any benefit of doing that right now vs doing an extra lookup in the store. Let me know if you think otherwise.

This will be used in prometheus metrics to tag a counter with the service name and link name. And also in logging when requests fail

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
asloobq: Its actually for both email and phone numbers. I copy pasted the description from the non-service metrics

src/main/java/com/uid2/operator/service/SecureLinkValidatorService.java
asloobq: Never mind. I changed it to fetch the service name

thomasm-ttd (approved):

PR #262, opened 2023-10-31
Do not log a "Cannot get a default keyset with SITE ID x" warning if /key/sharing participant does not have a keyset

cYKatherine:


src/main/java/com/uid2/operator/model/KeyManager.java
```diff
@@ -47,7 +47,6 @@ private Keyset getDefaultKeysetBySiteId(int siteId) {
         List<Keyset> keysets = this.keysetProvider.getSnapshot().getAllKeysets().values().stream()
                 .filter(s -> s.isEnabled() && s.isDefault() && s.getSiteId() == siteId).collect(Collectors.toList());
         if (keysets.isEmpty()) {
-            LOGGER.warn(String.format("Cannot get a default keyset with SITE ID %d.", siteId));
```
jon8787, line 50: I think we want to log in all cases except one specific scenario (ID_READER /key/sharing). So should we pass a `boolean logIfNotFound` into here instead?

jon8787 (approved):

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -569,6 +569,8 @@ public void handleKeysSharing(RoutingContext rc) {
             resp.put("master_keyset_id", MASTER_KEYSET_ID_FOR_SDKS);
             if (defaultKeyset != null) {
                 resp.put("default_keyset_id", defaultKeyset.getKeysetId());
+            } else if (!(roleAuthorize.hasRole(Role.SHARER))) {
```
jon8787, line 572: don't need those inner parentheses ()

PR #263, opened 2023-10-31
Fix enclavePlatform issue

cYKatherine:


src/main/java/com/uid2/operator/Main.java
```diff
@@ -491,16 +491,19 @@ private Map.Entry<UidCoreClient, UidOptOutClient> createUidClients(Vertx vertx,
     }
 
     private IOperatorKeyRetriever createOperatorKeyRetriever() throws Exception {
-        var enclavePlatform = this.config.getString("enclave_platform", "");
+        var enclavePlatform = this.config.getString("enclave_platform");
+        if (enclavePlatform == null || enclavePlatform.equals("")) {
```
vishalegbert-ttd, line 495: Perhaps isEmpty is better

src/main/java/com/uid2/operator/Main.java
```diff
@@ -491,16 +491,19 @@ private Map.Entry<UidCoreClient, UidOptOutClient> createUidClients(Vertx vertx,
...
         switch (enclavePlatform) {
             case "azure-cc": {
                 var vaultName = this.config.getString(Const.Config.AzureVaultNameProp);
                 var secretName = this.config.getString(Const.Config.AzureSecretNameProp);
                 return OperatorKeyRetrieverFactory.getAzureOperatorKeyRetriever(vaultName, secretName);
             }
             default: {
-                // default to load from config
-                return () -> this.config.getString(Const.Config.CoreApiTokenProp);
+                throw new IllegalArgumentException();
```
vishalegbert-ttd, line 506: I wonder if this is the right exception since this isn't actually an argument

We should add a message to this exception

src/main/java/com/uid2/operator/Main.java
```diff
@@ -491,16 +491,19 @@ private Map.Entry<UidCoreClient, UidOptOutClient> createUidClients(Vertx vertx,
     }
 
     private IOperatorKeyRetriever createOperatorKeyRetriever() throws Exception {
-        var enclavePlatform = this.config.getString("enclave_platform", "");
+        var enclavePlatform = this.config.getString("enclave_platform");
+        if (enclavePlatform == null || enclavePlatform.isEmpty()) {
```
yishi-ttd, line 495: nit: Strings.isNullOrEmpty()

yishi-ttd (approved):

vishalegbert-ttd (approved):

PR #264, opened 2023-11-01
Remove CSTG timestamp check

mcollins-ttd:


src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
```diff
@@ -33,7 +33,6 @@ public enum ResponseStatus {
         MissingParams,
         BadPublicKey,
         BadSubscriptionId,
-        BadTimestamp,
```
sunnywu, line 36: we can't remove this right otherwise the enum values of the errors below will be off by 1? so we can only leave it here forever?

sunnywu: @mcollins-ttd confirmed enum is only used as string so it's fine

sunnywu (approved):

PR #269, opened 2023-11-02
wait before error quit

yishi-ttd:
Wait a bit before error quit. Allow error log to be flushed to logs.

`((LoggerContext)org.slf4j.LoggerFactory.getILoggerFactory()).stop();` does not always work - it depends on OS buffer implementation.

This is a known issue: https://jira.qos.ch/browse/LOGBACK-735

Without this fix we will see weird log truncation in GCP. The real exception will not be logged - not friendly for touble shooting.

image

### DIFF
--- a/src/main/java/com/uid2/operator/Main.java
+++ b/src/main/java/com/uid2/operator/Main.java
@@ -194,12 +194,23 @@ public class Main {
                 return;
             }
 
+            boolean errorQuit = false;
+
             try {
                 Main app = new Main(vertx, ar.result());
                 app.run();
             } catch (Exception e) {
                 LOGGER.error("Error: " + e.getMessage(), e);
-                ((LoggerContext)org.slf4j.LoggerFactory.getILoggerFactory()).stop(); // flush logs before shutdown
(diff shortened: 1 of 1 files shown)


src/main/java/com/uid2/operator/Main.java
```diff
@@ -194,12 +194,23 @@ else if (!Utils.isProductionEnvironment()) {
...
                 LOGGER.error("Error: " + e.getMessage(), e);
-                ((LoggerContext)org.slf4j.LoggerFactory.getILoggerFactory()).stop(); // flush logs before shutdown
+                errorQuit = true;
+            }
+
+            if(errorQuit){
+                // allow log to be flushed before quit app.
+                try {
+                    Thread.sleep(1000);
+                } catch (InterruptedException e) {
```
thomasm-ttd, line 211: Is it worth catching all exceptions here?

yishi-ttd: updated

thomasm-ttd (approved):

PR #289, opened 2023-11-27
UID2-2112 Allow generating v4 tokens by percentage (fall back to default token version otherwise)

jon8787:
config name has changed from "advertising_token_v4" to "advertising_token_v4_percentage"


src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -150,7 +144,7 @@ private void setupConfig(JsonObject config) {
 
         config.put("identity_scope", getIdentityScope().toString());
         config.put("advertising_token_v3", getTokenVersion() == TokenVersion.V3);
-        config.put("advertising_token_v4", getTokenVersion() == TokenVersion.V4);
+        config.put("advertising_token_v4_percentage", getTokenVersion() == TokenVersion.V4 ? 100 : 0);
```
sunnywu, line 147: should we get as many unit tests to use v4 token AND v2 token as possible? (currently getTokenVersion will return V2), e.g. 2 sub classes of UIDOperatorVerticleTest 1 using v2 1 using v4

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -48,7 +47,8 @@ public class UIDOperatorService implements IUIDOperatorService {
     private final Duration refreshIdentityAfter;
 
     private final OperatorIdentity operatorIdentity;
-    private final TokenVersion advertisingTokenVersion;
+    private final TokenVersion tokenVersionToUseIfNotV4;
+    private final int advertisingTokenV4Percentage;
```
sunnywu, line 51: add a comment to say if we wanna disable v4, this needs to be set to 0?

sunnywu:
Maybe change the title to 

_UID2-2112 Allow generate v4 tokens by percentage (fall back to default token version otherwise)_ 


And add some top description to mention changes in unit tests and the actual config name change (advertising_token_v4 => advertising_token_v4_percentage)

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -295,16 +293,13 @@ private RefreshToken createRefreshToken(PublisherIdentity publisherIdentity, Use
...
 
     private AdvertisingToken createAdvertisingToken(PublisherIdentity publisherIdentity, UserIdentity userIdentity, Instant now) {
-        return new AdvertisingToken(
-                this.advertisingTokenVersion,
-                now,
-                now.plusMillis(identityExpiresAfter.toMillis()),
-                this.operatorIdentity,
-                publisherIdentity,
-                userIdentity);
+        int randomNum = ThreadLocalRandom.current().nextInt(1, 101);
```
sunnywu, line 296: is it worth _optimising_ and do this random num generation only when advertisingTokenV4Percentage >0?

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -295,16 +293,13 @@ private RefreshToken createRefreshToken(PublisherIdentity publisherIdentity, Use
...
+        var tokenVersion = (randomNum <= this.advertisingTokenV4Percentage) ? TokenVersion.V4 : this.tokenVersionToUseIfNotV4;
+
+        return new AdvertisingToken(tokenVersion, now, now.plusMillis(identityExpiresAfter.toMillis()), this.operatorIdentity, publisherIdentity, userIdentity);
     }
 
-    protected class GlobalOptoutResult {
+    static protected class GlobalOptoutResult {
```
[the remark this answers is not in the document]
jon8787: it's https://stackoverflow.com/questions/70324/java-inner-class-and-static-nested-class , but basically non-static means it can access fields of its enclosing class.

src/main/java/com/uid2/operator/service/UIDOperatorService.java
jon8787: no :)

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
jon8787: `UidOperatorVerticleV4Test` runs the entire suite again with token v4, is this what you meant?

sunnywu: no forgot about that, that's fine then

src/main/java/com/uid2/operator/service/UIDOperatorService.java
jon8787: added comment near the config line because it shows the actual config `advertising_token_v4_percentage`

conf/local-config.json
```diff
@@ -13,7 +13,7 @@
   "refresh_token_expires_after_seconds": 86400,
   "refresh_identity_token_after_seconds": 900,
   "advertising_token_v3": false,
-  "advertising_token_v4": false,
+  "advertising_token_v4_percentage": 0,
```
sunnywu, line 16: you probably wanna check if there are any other repos like e2e/validator that has specified this config in the config files like this and fix them first (and likewise for prod k8-deployment/euid-deployment) which is outside the scope of this PR.

jon8787: interestingly couldn't find the old config anywhere, other than in @mcollins-ttd's personal repo

mcollins-ttd (approved):

PR #293, opened 2023-11-29
Record advertising token version in uid2_token_response_status_count metric

mcollins-ttd:
This pull request adds the `advertising_token_version` label to the `uid2_token_response_status_count` metric.

It also removes the second layer of caching of the counter - Micrometer already does this for us. Having to update the `TokenResponseKey` class every time we add a label complicates things.

### DIFF
--- a/src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
+++ b/src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
@@ -2,6 +2,7 @@ package com.uid2.operator.monitoring;
 
 import com.uid2.operator.model.RefreshResponse;
 import com.uid2.operator.vertx.UIDOperatorVerticle;
+import com.uid2.shared.model.TokenVersion;
 import com.uid2.shared.store.ISiteStore;
 import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.Metrics;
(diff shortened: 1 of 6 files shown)


src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
```diff
@@ -42,68 +39,37 @@ public enum ResponseStatus {
         Unknown
     }
 
-    private static final Map<TokenResponseKey, Counter> TokenResponseCounters = new ConcurrentHashMap<>();
```
sunnywu, line 45: why is this not needed any more?

mcollins-ttd: Micrometer will return the same counter when your register with the same name and tags. This is just duplicating that functionality. I believe it was meant to be an optimisation but doubt it actually makes much of a difference.

jon8787 (approved):

src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
mcollins-ttd: It's also a maintenance burden: for every label added or removed you have to update this class accordingly.

sunnywu: i'll take this at face value 😅

sunnywu (approved):

PR #304, opened 2023-12-06
UID2-988 - Updated local Docker config to enable E2E interaction with optout

gmsdelmundo:


conf/local-e2e-docker-public-config.json
```diff
@@ -1,6 +1,6 @@
 {
   "service_instances": 1,
-  "storage_mock": true,
```
vishalegbert-ttd, line 3: How was this working before?

conf/local-e2e-docker-private-config.json
```diff
@@ -20,20 +20,13 @@
...
-  "optout_max_partitions": 30,
-  "optout_partition_interval": 86400,
   "client_side_token_generate": false,
-  "validate_service_links": false
+  "validate_service_links": false,
+  "optout_s3_bucket": "test-optout-bucket",
+  "optout_s3_folder": "optout-v2/",
+  "optout_metadata_path": "http://optout:8081/optout/refresh",
+  "optout_api_uri": "http://optout:8081/optout/replicate",
+  "optout_api_token": "OPLCLAjLRWcVlCDl9+BbwR38gzxYdiWFa751ynWLuI7JU4iA=",
```
vishalegbert-ttd, line 29: I thought we didn't need this anymore?

conf/local-e2e-docker-public-config.json
gmsdelmundo: Since `coreAttestUrl` is not null, most of operator's stores are already connected to core rather than S3

As such, the `storage_mock` boolean is only used to check whether the optout store should be mocked or connected to S3

conf/local-e2e-docker-private-config.json
vishalegbert-ttd: I think it should use a JWT from Core to authenticate with opt out now

gmsdelmundo: Oh yeah optout_api_token specifically is not needed, removed

vishalegbert-ttd (approved):

PR #308, opened 2023-12-08
UID2-2338 build test pipeline for operator

cYKatherine:
Sample build: :
* E2E test by itself: https://github.com/IABTechLab/uid2-operator/actions/runs/7191225936
* Together with "Publish Public Operator": https://github.com/IABTechLab/uid2-operator/actions/runs/7191241648

Review this together with: 
1. https://github.com/IABTechLab/uid2-e2e/pull/4
2. https://github.com/IABTechLab/uid2-shared-actions/pull/35
3. https://github.com/IABTechLab/uid2-operator/pull/308


.github/workflows/run-e2e-tests-on-operator.yaml
```diff
@@ -0,0 +1,23 @@
...
+    inputs:
+      operator_image_version:
+        description: If set, the version number will not be incremented and the given number will be used.
+        type: string
+        default: 'latest'
+
+jobs:
+  e2e-test:
+    name: E2E Test
+    uses: IABTechLab/uid2-shared-actions/.github/workflows/shared-run-e2e-tests.yaml@kcc-UID2-2338-build-test-pipeline-for-operator
```
cYKatherine, line 19: This will be removed and replaced by a tag after merging https://github.com/IABTechLab/uid2-shared-actions/pull/35/files

.github/workflows/run-e2e-tests-on-operator.yaml
```diff
@@ -0,0 +1,23 @@
+name: Run operator E2E tests
+on:
+  workflow_dispatch:
+    inputs:
+      operator_image_version:
+        description: If set, the version number will not be incremented and the given number will be used.
```
gmsdelmundo, line 6: Even if we don't set the image version, shouldn't the version number still be the same? i.e. not incremented

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-operator</artifactId>
-    <version>5.24.17-5b18819a00</version>
+    <version>5.24.59-SNAPSHOT</version>
```
gmsdelmundo, line 9: Don't forget to revert

.github/workflows/run-e2e-tests-on-operator.yaml
cYKatherine: This workflow is also used in "Publish public operator" (the above file `.github/workflows/publish-public-operator-docker-image.yaml`) In that workflow, it will use the new image version that Image step builds.

.github/workflows/run-e2e-tests-on-operator.yaml
```diff
@@ -0,0 +1,23 @@
...
+        type: string
+        default: 'latest'
+  workflow_call:
+    inputs:
+      operator_image_version:
+        description: If set, the version number will not be incremented and the given number will be used.
```
vishalegbert-ttd, line 12: Update

.github/workflows/run-e2e-tests-on-operator.yaml
```diff
@@ -0,0 +1,23 @@
+name: Run operator E2E tests
+on:
+  workflow_dispatch:
+    inputs:
```
vishalegbert-ttd, line 4: Optional params for core and opt out versions, default is latest

.github/workflows/run-e2e-tests-on-operator.yaml
cYKatherine: Updated and added core, optout and e2e image version as inputs as well

.github/workflows/run-e2e-tests-on-operator.yaml
```diff
@@ -0,0 +1,85 @@
...
```
cYKatherine, line 4: Was thinking whether I should pass in the environment variables here. But I guess we don't need that as those will be set for each individual workflow, shouldn't be typed in manually. @gmsdelmundo @vishalegbert-ttd

vishalegbert-ttd (approved):

PR #313, opened 2023-12-18
sch-UID2-2555-role-validation-for-service-links-in-operator

sophia-chen-ttd:
Added role validation for service links.

Tested
- Trying to validate a service link with the given role should return true
- Trying to validate a service link without the given role should return false


src/main/java/com/uid2/operator/service/SecureLinkValidatorService.java
```diff
@@ -45,6 +46,10 @@ public boolean validateRequest(RoutingContext rc, JsonObject requestJsonObject)
                         LOGGER.warn("ClientKey has ServiceId set, but LinkId in request was not authorized. ServiceId: {}, LinkId in request: {}", clientKey.getServiceId(), linkId);
                         return false;
                     }
+                    if (!serviceLink.getRoles().contains(role)) {
+                        LOGGER.warn("ServiceLink {} does not have have role {}", linkId, role);
```
vishalegbert-ttd, line 50: Maybe log the path from the rc if possible

atarassov-ttd (approved):

PR #321, opened 2024-01-02
Add `ENFORCE_HTTPS` environment variable

cYKatherine:
The reason why we need this change is because we are using `bore` to forward local network to a public endpoint and it's using http protocol. To override the value of `enforce_https`, we need to set `DEPLOYMENT_ENVIRONMENT` so that it will pick the correct config fire from here: https://github.com/IABTechLab/uid2-operator/blob/main/scripts/gcp-oidc/entrypoint.sh#L23. The E2E test used to use the integ config, I'm adding this e2e `DEPLOYMENT_ENVIRONMENT` so that we have `enforce_https` set to false for E2E tests.


scripts/gcp-oidc/entrypoint.sh
```diff
@@ -43,6 +43,11 @@ if [ -n "${CORE_BASE_URL}" -a -n "${OPTOUT_BASE_URL}" -a "${DEPLOYMENT_ENVIRONME
     sed -i "s#https://optout-integ.uidapi.com#${OPTOUT_BASE_URL}#g" ${FINAL_CONFIG}
 fi
 
+# -- replace `enforce_https` value to false if IS_E2E_TEST is true
+if [ "${IS_E2E_TEST}" == 'true' ]; then
```
yishi-ttd, line 47: I would prefer to rename the env name to "enforce_https" to be more specific.

Do you plan to have add more e2e tricks in future?

cYKatherine: Good point, fixed

yishi-ttd (approved):

PR #329, opened 2024-01-12
UID2-2340 Allow override for `enforce_https`

cYKatherine:
Please review in order of:

1. https://github.com/IABTechLab/uid2-operator/pull/329
2. https://github.com/IABTechLab/uid2-shared-actions/pull/50
3. https://github.com/IABTechLab/uid2-operator/pull/328

I’ve tested it working: 
- As an individual E2E task: https://github.com/IABTechLab/uid2-operator/actions/runs/7536390454/job/20513736679
- As a part of the publishing azure operator pipeline: https://github.com/IABTechLab/uid2-operator/actions/runs/7535436593

### DIFF
--- a/scripts/azure-cc/Dockerfile
+++ b/scripts/azure-cc/Dockerfile
@@ -27,7 +27,6 @@ RUN tar xzvf /app/static.tar.gz --no-same-owner --no-same-permissions && rm -f /
 COPY ./entrypoint.sh /app/
 RUN chmod a+x /app/entrypoint.sh
 
-RUN adduser -D uid2-operator && mkdir -p /opt/uid2 && chmod 777 -R /opt/uid2 && mkdir -p /app && chmod 705 -R /app && mkdir -p /app/file-uploads && chmod 777 -R /app/file-uploads
-USER uid2-operator
+RUN mkdir -p /opt/uid2 && chmod 777 -R /opt/uid2 && mkdir -p /app && chmod 705 -R /app && mkdir -p /app/file-uploads && chmod 777 -R /app/file-uploads
 
 CMD ["/app/entrypoint.sh"]
--- a/scripts/azure-cc/Dockerfile
+++ b/scripts/azure-cc/Dockerfile
@@ -27,6 +27,7 @@ RUN tar xzvf /app/static.tar.gz --no-same-owner --no-same-permissions && rm -f /
 COPY ./entrypoint.sh /app/
 RUN chmod a+x /app/entrypoint.sh
 
-RUN mkdir -p /opt/uid2 && chmod 777 -R /opt/uid2 && mkdir -p /app && chmod 705 -R /app && mkdir -p /app/file-uploads && chmod 777 -R /app/file-uploads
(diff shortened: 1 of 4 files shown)


scripts/azure-cc/Dockerfile
```diff
@@ -27,7 +27,6 @@ RUN tar xzvf /app/static.tar.gz --no-same-owner --no-same-permissions && rm -f /
 COPY ./entrypoint.sh /app/
 RUN chmod a+x /app/entrypoint.sh
 
-RUN adduser -D uid2-operator && mkdir -p /opt/uid2 && chmod 777 -R /opt/uid2 && mkdir -p /app && chmod 705 -R /app && mkdir -p /app/file-uploads && chmod 777 -R /app/file-uploads
-USER uid2-operator
+RUN mkdir -p /opt/uid2 && chmod 777 -R /opt/uid2 && mkdir -p /app && chmod 705 -R /app && mkdir -p /app/file-uploads && chmod 777 -R /app/file-uploads
```
cYKatherine, line 30: We need this change because with the uid2-operator user it doesn’t have 777 access to the /tmp folder. Remove User to be consistent with gcp-oidc: https://github.com/IABTechLab/uid2-operator/blob/main/scripts/gcp-oidc/Dockerfile#L32

yishi-ttd (approved):

scripts/azure-cc/Dockerfile
yishi-ttd: Hold on.. I remember we remove user in GCP for some specific reason: not be able to read token.
If possible let's keep the uid2-operator user.
Can we update jq_inplace_update_json function to put tmp file in same folder with final_config where we should have write permission? ( See comment below )

scripts/azure-cc/entrypoint.sh
```diff
@@ -2,6 +2,16 @@
...
 # This script must be compatible with Ash (provided in eclipse-temurin Docker image) and Bash
 
+# for number/boolean
+# https://jqlang.github.io/jq/manual/
+# --argjson foo 123 will bind $foo to 123.
+function jq_inplace_update_json() {
+    local file=$1
+    local field=$2
+    local value=$3
+    jq --argjson v "$value" ".$field = \$v" "$file" > tmp.json && mv tmp.json "$file"
```
yishi-ttd, line 12: Can it work if we define a local `$tmp = "$file.tmp", and replace `tmp.json` with it?

scripts/azure-cc/entrypoint.sh
```diff
@@ -2,6 +2,18 @@
...
+TMP_FINAL_CONFIG="/tmp/final-config.tmp"
+
+# for number/boolean
+# https://jqlang.github.io/jq/manual/
+# --argjson foo 123 will bind $foo to 123.
+function jq_inplace_update_json() {
+    local file=$1
+    local field=$2
+    local value=$3
+    jq --argjson v "$value" ".$field = \$v" "$file" > $TMP_FINAL_CONFIG && mv $TMP_FINAL_CONFIG "$file"
```
yishi-ttd, line 14: better to move the TMP_FINAL_CONFIG path inside this function.

PR #331, opened 2024-01-16
Move TMP_FINAL_CONFIG into the jq_inplace_update_json definition

cYKatherine:


scripts/azure-cc/entrypoint.sh
```diff
@@ -2,12 +2,11 @@
 #
 # This script must be compatible with Ash (provided in eclipse-temurin Docker image) and Bash
 
-TMP_FINAL_CONFIG="/tmp/final-config.tmp"
-
 # for number/boolean
 # https://jqlang.github.io/jq/manual/
 # --argjson foo 123 will bind $foo to 123.
 function jq_inplace_update_json() {
+    TMP_FINAL_CONFIG="/tmp/final-config.tmp"
```
yishi-ttd, line 9: 1. I would add a `local` modifier to make it only accessible inside this function.
2. I would rename the name to be more specific, from `TMP_FINAL_CONFIG="/tmp/final-config.tmp"` to `local TMP_FILE="/tmp/json.tmp"`

PR #346, opened 2024-02-09
Updated AWS private operator scripts for E2E test pipeline

gmsdelmundo:


scripts/aws/UID_CloudFormation.template.yml
```diff
@@ -2,7 +2,7 @@ AWSTemplateFormatVersion: 2010-09-09
 Description: UID 2.0 CloudFormation template
 Parameters:
   APIToken:
-    Description: UID API Token
+    Description: UID2 API Token
```
thomasm-ttd, line 5: There is a screen shot with this name in the docs on UnifiedId.com. I suggest we don't change it now.

scripts/aws/UID_CloudFormation.template.yml
thomasm-ttd, line 1: We should make the same changes in the EUID Cloud Formation Template

scripts/aws/entrypoint.sh
```diff
@@ -1,68 +1,122 @@
...
+echo "Loading env vars via proxy..."
+
+if [ "${IDENTITY_SCOPE}" = "UID2" ]; then
+  USER_DATA=$(curl -s -x socks5h://127.0.0.1:3305 http://169.254.169.254/latest/user-data)
+  UID2_CONFIG_SECRET_KEY=$([[ "$(echo "${USER_DATA}" | grep UID2_CONFIG_SECRET_KEY=)" =~ ^export\ UID2_CONFIG_SECRET_KEY=\"(.*)\"$ ]] && echo "${BASH_REMATCH[1]}" || echo "uid2-operator-config-key")
+  CORE_BASE_URL=$([[ "$(echo "${USER_DATA}" | grep CORE_BASE_URL=)" =~ ^export\ CORE_BASE_URL=\"(.*)\"$ ]] && echo "${BASH_REMATCH[1]}" || echo "")
+  OPTOUT_BASE_URL=$([[ "$(echo "${USER_DATA}" | grep OPTOUT_BASE_URL=)" =~ ^export\ OPTOUT_BASE_URL=\"(.*)\"$ ]] && echo "${BASH_REMATCH[1]}" || echo "")
+elif [ "${IDENTITY_SCOPE}" = "EUID" ]; then
+  USER_DATA=$(curl -s -x socks5h://127.0.0.1:3305 http://169.254.169.254/latest/user-data)
+  UID2_CONFIG_SECRET_KEY=$([[ "$(echo "${USER_DATA}" | grep UID2_CONFIG_SECRET_KEY=)" =~ ^export\ UID2_CONFIG_SECRET_KEY=\"(.*)\"$ ]] && echo "${BASH_REMATCH[1]}" || echo "uid2-operator-config-key")
```
thomasm-ttd, line 24: This line looks wrong. The old code used EUID, you have UID2. Are you sure this is correct?

scripts/aws/entrypoint.sh
```diff
@@ -1,68 +1,122 @@
...
+  CORE_BASE_URL=$([[ "$(echo "${USER_DATA}" | grep CORE_BASE_URL=)" =~ ^export\ CORE_BASE_URL=\"(.*)\"$ ]] && echo "${BASH_REMATCH[1]}" || echo "")
```
thomasm-ttd, line 25: The core and optout URL commands don't need to be inside the identity scope if

scripts/aws/entrypoint.sh
gmsdelmundo: Updated it here: https://github.com/IABTechLab/uid2-operator/pull/405

scripts/aws/UID_CloudFormation.template.yml
gmsdelmundo: This is only for UID2

EUID PR is here: https://github.com/IABTechLab/uid2-operator/pull/405

.github/workflows/run-e2e-tests-on-operator.yaml
```diff
@@ -1,85 +1,103 @@
...
-        default: main
-      admin_branch:
-        description: 'Admin: use this branch for Admin config to test with'
-        type: string
-        default: main
-      operator_branch:
-        description: 'Operator: use this branch for Operator config to test with'
-        type: string
-        default: main
+        - aws
```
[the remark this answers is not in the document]
gmsdelmundo: There can only be a maximum of 10 input parameters for a GitHub pipeline, so the image/branch/AWS args were changed to JSON inputs

scripts/aws/entrypoint.sh
```diff
@@ -1,68 +1,126 @@
...
 get_config_value() {
-  jq -r ".\"$1\"" /app/conf/config-final.json
+  jq -r ".\"$1\"" ${FINAL_CONFIG}
+}
+
+# for number/boolean
+# https://jqlang.github.io/jq/manual/
+# --argjson foo 123 will bind $foo to 123.
+TMP_FINAL_CONFIG="/tmp/final-config.tmp"
+function jq_inplace_update_json() {
```
thomasm-ttd, line 80: Where is this used?

thomasm-ttd (approved):

scripts/aws/entrypoint.sh
gmsdelmundo: I think this might've been a leftover of the previous enforce_https override

I'll be removing this

.github/workflows/run-e2e-tests-on-operator.yaml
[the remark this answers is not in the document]
gmsdelmundo: Fixed this - since the pipeline has always been using the pipeline's branch as the operator config branch rather than the input param, everything except the AWS arg list should fit into the original input arg format

PR #348, opened 2024-02-12
Add `skip_e2e_test` as inputs

cYKatherine:


.github/workflows/publish-all-operators.yaml
```diff
@@ -10,6 +10,11 @@ on:
         - Major
         - Minor
         - Patch
+      skip_e2e_test:
+        description: If true, will skip the step for E2E tests
+        required: true
+        type: boolean
+        default: true
```
thomasm-ttd, line 17: Lets make the default false

.github/workflows/publish-all-operators.yaml
```diff
@@ -10,6 +10,11 @@ on:
         - Major
         - Minor
         - Patch
+      skip_e2e_test:
+        description: If true, will skip the step for E2E tests
+        required: true
```
thomasm-ttd, line 15: This is not required if you give a default value

.github/workflows/publish-azure-cc-enclave-docker.yaml
```diff
@@ -15,6 +15,11 @@ on:
         description: If set, the version number will not be incremented and the given number will be used.
         type: string
         default: ''
+      skip_e2e_test:
+        description: If true, will skip the step for E2E tests
+        required: true
+        type: boolean
+        default: true
```
thomasm-ttd, line 22: default: false

PR #356, opened 2024-02-19
Remove the enforce_https setting

thomasm-ttd:
Updating shared to pass attestation url as part of the attestation document


cYKatherine (approved):

scripts/gcp-oidc/entrypoint.sh
```diff
@@ -2,17 +2,6 @@
 #
 # This script must be compatible with Ash (provided in eclipse-temurin Docker image) and Bash
 
-# for number/boolean
```
cYKatherine, line 5: These are actually not used, we will need to update those in shared-actions.

PR #363, opened 2024-02-26
Use the returned OptOut URL from core

thomasm-ttd:
Renamed AttestationTokenRetriever to AttestationResponseHandler as it does more than just retrieve the token.

All unit tests pass.
Tested with local e2e.


conf/local-e2e-docker-public-config.json
```diff
@@ -27,7 +27,7 @@
   "validate_service_links": true,
   "optout_s3_bucket": "test-optout-bucket",
   "optout_s3_folder": "optout-v2/",
-  "optout_metadata_path": "http://optout:8081/optout/refresh",
+  "optout_metadata_path": "/optout/refresh",
```
cYKatherine, line 30: Why do we want to remove the http here?

thomasm-ttd: The base URL is what is returned from Core, so we only need the relative path in config.

cYKatherine: Mmm then why do we keep the base in `conf/validator-latest-e2e-docker-public-config.json` below

conf/validator-latest-e2e-docker-public-config.json
```diff
@@ -0,0 +1,35 @@
...
+  "identity_v3": false,
+  "identity_scope": "uid2",
+  "enable_v2_encryption": true,
+  "client_side_token_generate": true,
+  "client_side_token_generate_domain_name_check_enabled": true,
+  "key_sharing_endpoint_provide_site_domain_names": true,
+  "validate_service_links": true,
+  "optout_s3_bucket": "test-optout-bucket",
+  "optout_s3_folder": "optout-v2/",
+  "optout_metadata_path": "http://optout:8081/optout/refresh",
```
cYKatherine, line 31: Should we remove the base here?

conf/local-e2e-docker-public-config.json
thomasm-ttd: Because that is the version that is running in Prod at the moment. It expects to have the full URL. Once this version becomes the latest, that file will need to be updated.

conf/validator-latest-e2e-docker-public-config.json
thomasm-ttd: No, I had to add this file specifically for that setting as it is for the previous version of operator which expects the full path

cYKatherine (approved):

conf/validator-latest-e2e-docker-public-config.json
cYKatherine: Should we add a comment to the top of this config file, so that later on when all the previous version of operators are gone we can delete this file?

cYKatherine: No comments in json as discussed offline :( hopefully this conversation thread will help us recall what happened in this ticket 🤞

PR #364, opened 2024-02-28
UID2-2558 Use commit_pr_and_merge workflow

cYKatherine:


.github/workflows/publish-all-operators.yaml
```diff
@@ -57,13 +58,19 @@ jobs:
           echo "Version number updated from $current_version to $new_version"
 
       - name: Commit pom.xml and version.json
-        uses: EndBug/add-and-commit@v9
+        if: ${{ steps.checkRelease.outputs.is_release != 'true' }}
```
cYKatherine, line 61: actually we don't need this as we won't deploy snapshot version in this workflow. Removing it.

thomasm-ttd (approved):

thomasm-ttd (approved):

PR #372, opened 2024-02-28
Syw UID2-2876 dsp check operator key sharing change

sunnywu:


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -106,6 +106,7 @@ public class UIDOperatorVerticle extends AbstractVerticle {
     public final static int MASTER_KEYSET_ID_FOR_SDKS = 9999999; //this is because SDKs have an issue where they assume keyset ids are always positive; that will be fixed.
     public final static long OPT_OUT_CHECK_CUTOFF_DATE = Instant.parse("2023-09-01T00:00:00.00Z").getEpochSecond();
 
+    protected int maxSharingLifetimeSeconds;
```
sunnywu, line 109: for easier unit testing (when we want to set custom value), this variable it created and can be overwritten on a per unit test basis see  keySharingKeysets_SHARER_CustomMaxSharingLifetimeSeconds test

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -534,7 +536,13 @@ public void handleKeysSharing(RoutingContext rc) {
             } else if (roleAuthorize.hasRole(Role.SHARER)) {
                 LOGGER.warn(String.format("Cannot get a default keyset with SITE ID %d. Caller will not be able to encrypt tokens..", clientKey.getSiteId()));
             }
+            // this is written out as a String, i.e. in the JSON response of key/sharing endpoint, it would show: 
+            // "token_expiry_seconds" : "2592000"
+            // it should be an integer instead, but we can't change it until we confirm that the oldest version of each of our SDKs support this
             resp.put("token_expiry_seconds", getSharingTokenExpirySeconds());
+            resp.put("max_sharing_lifetime_seconds", maxSharingLifetimeSeconds);
```
mcollins-ttd, line 543: Should this only be present if the role is SHARER? cc @jon8787

jon8787: Yep that's right

PR #373, opened 2024-02-29
Rename feature branch input helper text

cYKatherine:


.github/workflows/run-e2e-tests-on-operator.yaml
```diff
@@ -13,33 +13,33 @@ on:
...
-        default: 'latest'
+        default: latest
       e2e_image_version:
         description: The pipeline will run the E2E test with this e2e image version
         type: string
-        default: 'latest'
+        default: latest
       core_branch:
-        description: 'The branch of UID2-core to test on'
+        description: The branch of UID2-core config to test on
```
thomasm-ttd, line 30: Might be clearer to say:
description: Core: use this branch for Core config to test with

Don't think we need to include the UID2 as that is implied

thomasm-ttd (approved):

PR #395, opened 2024-03-05
UID2-2165 Log origin field for InvalidHttpOrigin errors

caroline-ttd:
The logs look like: 
`InvalidHttpOrigin: site MegaTest Site (123): http://localhosk345678ljkhbjklnjkbjjkkjn.com, http://localhosk345678ljkhbjklnjkbjn.com, http://localhosk345n.com | site TestCorp Site (124): http://localhos.com, http://localhosk.com, http://localhoskl.com`

### DIFF
--- a/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
+++ b/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
@@ -5,6 +5,7 @@ import com.uid2.operator.model.*;
 import com.uid2.operator.model.IdentityScope;
 import com.uid2.operator.monitoring.IStatsCollectorQueue;
 import com.uid2.operator.monitoring.StatsCollectorHandler;
+import com.uid2.operator.monitoring.StatsCollectorVerticle;
 import com.uid2.operator.monitoring.TokenResponseStatsCollector;
 import com.uid2.operator.privacy.tcf.TransparentConsent;
 import com.uid2.operator.privacy.tcf.TransparentConsentParseResult;
@@ -50,6 +51,7 @@ import io.vertx.ext.web.RoutingContext;
 import io.vertx.ext.web.handler.BodyHandler;
 import io.vertx.ext.web.handler.CorsHandler;
 import io.vertx.ext.web.handler.StaticHandler;
+import lombok.extern.flogger.Flogger;
 import org.apache.http.HttpStatus;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -103,10 +105,14 @@ public class UIDOperatorVerticle extends AbstractVerticle {
     private final KeyManager keyManager;
     private final SecureLinkValidatorService secureLinkValidatorService;
     private final boolean cstgDoDomainNameCheck;
+    private final Duration invalidOriginProcessTimeInterval;
+    private final static long invalidOriginPerMin = 3600L; // We set the time interval to one hour
     public final static int MASTER_KEYSET_ID_FOR_SDKS = 9999999; //this is because SDKs have an issue where they assume keyset ids are always positive; that will be fixed.
     public final static long OPT_OUT_CHECK_CUTOFF_DATE = Instant.parse("2023-09-01T00:00:00.00Z").getEpochSecond();
 
     protected boolean keySharingEndpointProvideSiteDomainNames;
+    protected Map<Integer, Set<String>> invalidOriginMap;
+    private Instant lastInvalidOriginProcessTime;
 
     public UIDOperatorVerticle(JsonObject config,
                                boolean clientSideTokenGenerate,
@@ -144,6 +150,9 @@ public class UIDOperatorVerticle extends AbstractVerticle {
         this.keySharingEndpointProvideSiteDomainNames = config.getBoolean("key_sharing_endpoint_provide_site_domain_names", false);
         this._statsCollectorQueue = statsCollectorQueue;
         this.clientKeyProvider = clientKeyProvider;
+        this.invalidOriginProcessTimeInterval = Duration.ofMinutes(invalidOriginPerMin);
+        this.lastInvalidOriginProcessTime = Instant.now();
+        this.invalidOriginMap = new HashMap<>();
     }
 
     @Override
@@ -309,6 +318,7 @@ public class UIDOperatorVerticle extends AbstractVerticle {
 
             boolean allowedDomain = DomainNameCheckUtil.isDomainNameAllowed(origin, domainNames);
             if (!allowedDomain) {
+                handleInvalidHttpOriginError(clientSideKeypair.getSiteId(), origin);
                 SendClientErrorResponseAndRecordStats(ResponseStatus.InvalidHttpOrigin, 403, rc, "unexpected http origin", clientSideKeypair.getSiteId(), TokenResponseStatsCollector.Endpoint.ClientSideTokenGenerateV2, TokenResponseStatsCollector.ResponseStatus.InvalidHttpOrigin, siteProvider);
                 return;
             }
@@ -1756,6 +1766,18 @@ public class UIDOperatorVerticle extends AbstractVerticle {
     private void sendJsonResponse(RoutingContext rc, JsonArray json) {
         rc.response().putHeader(HttpHeaders.CONTENT_TYPE, "application/json")
                 .end(json.encode());
+    }
+
+    private void handleInvalidHttpOriginError(int siteId, String origin) {
+        Set<String> uniqueInvalidOrigins = invalidOriginMap.getOrDefault(siteId, new HashSet<>());
+        uniqueInvalidOrigins.add(origin);
+        invalidOriginMap.put(siteId, uniqueInvalidOrigins);
+
+        if (Duration.between(lastInvalidOriginProcessTime, Instant.now()).compareTo(invalidOriginProcessTimeInterval) >= 0) {
+            lastInvalidOriginProcessTime = Instant.now();
+            LOGGER.error("Invalid origin: " + invalidOriginMap.toString());
+            invalidOriginMap.clear();
+        }
     }
 
     public enum UserConsentStatus {
--- a/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
+++ b/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
@@ -1781,22 +1781,25 @@ public class UIDOperatorVerticle extends AbstractVerticle {
     private String generateInvalidHttpOriginMessage(Map<Integer, Set<String>> siteIdToInvalidOrigins) {
         StringBuilder invalidHttpOriginMessage = new StringBuilder();
         invalidHttpOriginMessage.append("InvalidHttpOrigin: ");
+        boolean mapHasFirstEle = false;
         for (Map.Entry<Integer, Set<String>> entry : siteIdToInvalidOrigins.entrySet()) {
+            if(mapHasFirstEle) {
+                invalidHttpOriginMessage.append(" | ");
+            }
+            mapHasFirstEle = true;
             int siteId = entry.getKey();
             Set<String> origins = entry.getValue();
             String siteName = getSiteName(siteProvider, siteId);
             String site = "site " + siteName + " (" + siteId + "): ";
             invalidHttpOriginMessage.append(site);
+            boolean setHasFirstEle = false;
             for (String origin : origins) {
-                invalidHttpOriginMessage.append(origin).append(", ");
+                if(setHasFirstEle) {
+                    invalidHttpOriginMessage.append(", ");
+                }
+                setHasFirstEle = true;
+                invalidHttpOriginMessage.append(origin);
             }
-            if (!origins.isEmpty()) {
-                invalidHttpOriginMessage.delete(invalidHttpOriginMessage.length() - 2, invalidHttpOriginMessage.length());
-            }
-            invalidHttpOriginMessage.append(" | ");
-        }
-        if (!siteIdToInvalidOrigins.isEmpty()) {
-            invalidHttpOriginMessage.delete(invalidHttpOriginMessage.length() - 3, invalidHttpOriginMessage.length());
         }
         return invalidHttpOriginMessage.toString();
     }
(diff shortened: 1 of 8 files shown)


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -50,6 +51,7 @@
 import io.vertx.ext.web.handler.BodyHandler;
 import io.vertx.ext.web.handler.CorsHandler;
 import io.vertx.ext.web.handler.StaticHandler;
+import lombok.extern.flogger.Flogger;
```
jon8787, line 54: Is this actually being used? I'm not sure if we've approved Lombok for use in Operator yet

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -5,6 +5,7 @@
 import com.uid2.operator.model.IdentityScope;
 import com.uid2.operator.monitoring.IStatsCollectorQueue;
 import com.uid2.operator.monitoring.StatsCollectorHandler;
+import com.uid2.operator.monitoring.StatsCollectorVerticle;
```
jon8787, line 8: are we using this?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -103,10 +105,14 @@ public class UIDOperatorVerticle extends AbstractVerticle {
     private final KeyManager keyManager;
     private final SecureLinkValidatorService secureLinkValidatorService;
     private final boolean cstgDoDomainNameCheck;
+    private final Duration invalidOriginProcessTimeInterval;
+    private final static long invalidOriginPerMin = 3600L; // We set the time interval to one hour
     public final static int MASTER_KEYSET_ID_FOR_SDKS = 9999999; //this is because SDKs have an issue where they assume keyset ids are always positive; that will be fixed.
     public final static long OPT_OUT_CHECK_CUTOFF_DATE = Instant.parse("2023-09-01T00:00:00.00Z").getEpochSecond();
 
     protected boolean keySharingEndpointProvideSiteDomainNames;
+    protected Map<Integer, Set<String>> invalidOriginMap;
```
jon8787, line 114: 1) should this be private?
2) We like to name maps `keyToValue` to make it clearer what the key is, so something like `siteIdToInvalidOrigins` ?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -103,10 +105,14 @@ public class UIDOperatorVerticle extends AbstractVerticle {
     private final KeyManager keyManager;
     private final SecureLinkValidatorService secureLinkValidatorService;
     private final boolean cstgDoDomainNameCheck;
+    private final Duration invalidOriginProcessTimeInterval;
+    private final static long invalidOriginPerMin = 3600L; // We set the time interval to one hour
```
jon8787, line 109: no need to have this const if it's only being used once

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1758,6 +1768,18 @@ private void sendJsonResponse(RoutingContext rc, JsonArray json) {
                 .end(json.encode());
     }
 
+    private void handleInvalidHttpOriginError(int siteId, String origin) {
+        Set<String> uniqueInvalidOrigins = invalidOriginMap.getOrDefault(siteId, new HashSet<>());
```
jon8787, line 1772: does `computeIfAbsent` work here?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1758,6 +1768,18 @@ private void sendJsonResponse(RoutingContext rc, JsonArray json) {
...
+        uniqueInvalidOrigins.add(origin);
+        invalidOriginMap.put(siteId, uniqueInvalidOrigins);
+
+        if (Duration.between(lastInvalidOriginProcessTime, Instant.now()).compareTo(invalidOriginProcessTimeInterval) >= 0) {
+            lastInvalidOriginProcessTime = Instant.now();
+            LOGGER.error("Invalid origin: " + invalidOriginMap.toString());
```
jon8787, line 1778: Curious, could you show how this looks when there are a few site ids and a few invalid origins in each?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1758,6 +1768,18 @@ private void sendJsonResponse(RoutingContext rc, JsonArray json) {
                 .end(json.encode());
     }
 
+    private void handleInvalidHttpOriginError(int siteId, String origin) {
```
jon8787, line 1771: will we add a feature flag check here?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: My bad, removed it.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: No, we did not use it. Removed it. It may be added when I did some tests locally and forget to remove.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: Yeah, it should be private. And will change the name.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: Yeah, I plan to. Will add it in the next revision.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: Yeah. It shows like "{123=[http://localhoskjlkjt.com, http://localhost.com]}". I plan to polish it to a better formed message.

jon8787: that's only 1 site id :)
Can we make it show site name?

caroline-ttd: Hi Jonathan, I like your idea for showing the site name which is more readable, does every site always have site name? And is the site name is also unique as site id?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: Yeah, it also works. I found our code base is using "computeIfAbsent". I will align with that and change it.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: Discussed offline, Site name is a mandatory field and we will use the format: `InvalidHttpOrigin: site ABC (123): [google.com](http://google.com/), [xyz.com](http://xyz.com/) | site DEF (456): [abc.com](http://abc.com/), [def.com](http://def.com/) | ... ` for the message.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1758,6 +1767,34 @@ private void sendJsonResponse(RoutingContext rc, JsonArray json) {
...
+            lastInvalidOriginProcessTime = Instant.now();
+            LOGGER.error(generateInvalidHttpOriginMessage(siteIdToInvalidOrigins));
+            siteIdToInvalidOrigins.clear();
+        }
+    }
+
+    private String generateInvalidHttpOriginMessage(Map<Integer, Set<String>> siteIdToInvalidOrigins) {
+        StringBuilder invalidHttpOriginMessage = new StringBuilder();
+        invalidHttpOriginMessage.append("InvalidHttpOrigin: ");
+        siteIdToInvalidOrigins.forEach((siteId, origins) -> {
```
jon8787, line 1784: better to use normal for-each loop https://stackoverflow.com/questions/16635398/java-8-iterable-foreach-vs-foreach-loop

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1758,6 +1767,34 @@ private void sendJsonResponse(RoutingContext rc, JsonArray json) {
...
+            String siteName = getSiteName(siteProvider, siteId);
+            String site = "site " + siteName + " (" + siteId + "): ";
+            invalidHttpOriginMessage.append(site);
+            origins.forEach(origin ->  invalidHttpOriginMessage.append(origin).append(", "));
```
jon8787, line 1788: for each loop

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1758,6 +1767,34 @@ private void sendJsonResponse(RoutingContext rc, JsonArray json) {
...
+            if (!origins.isEmpty()) {
+                invalidHttpOriginMessage.delete(invalidHttpOriginMessage.length() - 2, invalidHttpOriginMessage.length());
+            }
+            invalidHttpOriginMessage.append(" | ");
+        });
+        invalidHttpOriginMessage.delete(invalidHttpOriginMessage.length() - 3, invalidHttpOriginMessage.length());
```
jon8787, line 1794: is there a way to avoid needing to delete?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1758,6 +1767,34 @@ private void sendJsonResponse(RoutingContext rc, JsonArray json) {
                 .end(json.encode());
     }
 
+    private void handleInvalidHttpOriginError(int siteId, String origin) {
+        Set<String> uniqueInvalidOrigins = siteIdToInvalidOrigins.computeIfAbsent(siteId, k -> new HashSet<>());
+        uniqueInvalidOrigins.add(origin);
+
+        if (Duration.between(lastInvalidOriginProcessTime, Instant.now()).compareTo(Duration.ofMinutes(1)) >= 0) {
```
sunnywu, line 1774: log invalid domains every 1 minute?? 
I thought it's hourly at minimum or it's set by a config?

caroline-ttd: It should be one hour, I will update it.

sunnywu:
General comment - you should add the jira ticket name on the title of the PR

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: One way I can come up with is to use iterator and Map.Entry. And check whether it.hasNext is true to append the " | ". I feel those options are vert similar and delete looks more clear.

https://stackoverflow.com/questions/46898/how-do-i-efficiently-iterate-over-each-entry-in-a-java-map

sunnywu: just my 2cents - keep it simple and having extra " | "at the end is fine (remove the delete call), such formatting doesn't matter for our current goal :)

jon8787: `invalidHttpOriginMessage.append(" | ");` should be at the beginning of the loop, and only if the string is empty

sunnywu: _and only if the string is **NOT** empty_

?

caroline-ttd: In the beginning, it will look like this? `InvalidHttpOrigin: | site ABC (123): 123.com, 345.com | site DEF (456): 789.com, 435.com` 
Do we expect this? `InvalidHttpOrigin:  site ABC (123): 123.com, 345.com | site DEF (456): 789.com, 435.com | `

jon8787: per discussion, we can either add "InvalidHttpOrigin: " prefix after the loop, or use a boolean, or check for certain length

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -144,6 +147,9 @@ public UIDOperatorVerticle(JsonObject config,
         this.keySharingEndpointProvideSiteDomainNames = config.getBoolean("key_sharing_endpoint_provide_site_domain_names", false);
         this._statsCollectorQueue = statsCollectorQueue;
         this.clientKeyProvider = clientKeyProvider;
+        this.lastInvalidOriginProcessTime = Instant.now();
+        this.siteIdToInvalidOrigins = new HashMap<>();
```
jon8787, line 151: feel free to initialize the member declaration itself

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1758,6 +1767,43 @@ private void sendJsonResponse(RoutingContext rc, JsonArray json) {
...
+            lastInvalidOriginProcessTime = Instant.now();
+            LOGGER.error(generateInvalidHttpOriginMessage(siteIdToInvalidOrigins));
+            siteIdToInvalidOrigins.clear();
+        }
+    }
+
+    private String generateInvalidHttpOriginMessage(Map<Integer, Set<String>> siteIdToInvalidOrigins) {
+        StringBuilder invalidHttpOriginMessage = new StringBuilder();
+        invalidHttpOriginMessage.append("InvalidHttpOrigin: ");
+        boolean mapHasFirstEle = false;
```
jon8787, line 1784: it's ok to have long variable names, better not to cut off words

jon8787 (approved):

PR #403, opened 2024-03-07
Add /key/bidstream endpoint

mcollins-ttd:
This pull request adds the new `/key/bidstream` endpoint and includes changes to the `/key/sharing` endpoint. The `/key/bidstream` endpoint requires the `ID_READER` role and behaves similarly to the `/key/sharing` endpoint.

## Header fields

`/key/bidstream` has the following response header fields:
- `allow_clock_skew_seconds`
- `identity_scope`
- `max_bidstream_lifetime_seconds`

`/key/sharing` has the following response header fields:
- `allow_clock_skew_seconds` (new)
- `caller_site_id`
- `default_keyset_id`
- `identity_scope` (new)
- `master_keyset_id`
- `max_sharing_lifetime_seconds` (new) (if the client has the `SHARER` role)
- `token_expiry_seconds`

## Key fields
`/key/bidstream` keys do not contain `keyset_id`.


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -508,103 +520,143 @@ private String getSharingTokenExpirySeconds() {
...
+        addAllowClockSkewSecondsField(resp);
+    }
+
+    private void addSites(JsonObject resp, List<KeysetKey> keys, Map<Integer, Keyset> keysetMap) {
+        final List<Site> sites = getSitesWithDomainNames(keys, keysetMap);
+        if (sites != null) {
+            final List<JsonObject> sitesJson = sites.stream()
+                    .map(UIDOperatorVerticle::toJson)
+                    .collect(Collectors.toList());
+            resp.put("site_data", sitesJson);
```
sunnywu, line 593: the old version will always have site_data field (albeit an empty array), do we want to change this behaviour? (how does current/new c# sdk handle/assume?)

src/main/java/com/uid2/operator/Const.java
```diff
@@ -13,6 +13,9 @@ public class Config extends com.uid2.shared.Const.Config {
         public static final String FailureShutdownWaitHoursProp = "failure_shutdown_wait_hours";
         public static final String AllowLegacyAPIProp = "allow_legacy_api";
         public static final String SharingTokenExpiryProp = "sharing_token_expiry_seconds";
+        public static final String MaxBidstreamLifetimeSecondsProp = "max_bidstream_lifetime_seconds";
```
jon8787, line 16: Can we remove these consts? They're only used once so it's easier for the reader if they're inline

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -508,103 +520,143 @@ private String getSharingTokenExpirySeconds() {
...
+            final JsonArray keys = new JsonArray();
+            for (KeysetKey key : accessibleKeys) {
+                JsonObject keyObj = toJson(key);
                 Keyset keyset = keysetMap.get(key.getKeysetId());
 
-                if (keyset == null || !keyset.isEnabled()) {
-                    continue;
-                } else if (clientKey.getSiteId() == keyset.getSiteId()) {
+                // include 'keyset_id' field, if:
+                //   (a) a key belongs to caller's enabled site
```
jon8787, line 539: what do we mean by `enabled` site? I assume if a caller's site was disabled they wouldn't get here?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -4170,17 +4230,36 @@ void keySharingRotatingKeysets_IDREADER(String testRun, Vertx vertx, VertxTestCo
...
+        assertEquals(this.getIdentityScope().toString(), body.getString("identity_scope"));
+        assertNotNull(body.getInteger("allow_clock_skew_seconds"));
+
+        switch (endpoint) {
+            case SHARING:
+                assertEquals(clientSiteId, body.getInteger("caller_site_id"));
+                assertEquals(UIDOperatorVerticle.MASTER_KEYSET_ID_FOR_SDKS, body.getInteger("master_keyset_id"));
+                assertEquals(4, body.getInteger("default_keyset_id"));
+                // NOTE: this is intentionally a string, not an integer. See comment in UIDOperatorVerticle.
+                assertNotNull(body.getString("token_expiry_seconds"));
```
sunnywu, line 4255: do we wanna check assertNull(body.getInteger("max_bidstream_lifetime_seconds")) ?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -4170,17 +4230,36 @@ void keySharingRotatingKeysets_IDREADER(String testRun, Vertx vertx, VertxTestCo
...
+                break;
+            case BIDSTREAM:
+                assertNotNull(body.getInteger("max_bidstream_lifetime_seconds"));
```
sunnywu, line 4258: do we wanna check the value against an expected value?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -4170,17 +4230,36 @@ void keySharingRotatingKeysets_IDREADER(String testRun, Vertx vertx, VertxTestCo
...
+            checkSharingResponseHeaderFields(endpoint, body, clientSiteId);
+
+            checkEncryptionKeys(respJson, endpoint, clientSiteId, expectedKeys.toArray(new KeysetKey[0]));
             testContext.completeNow();
         });
     }
 
+    private void checkSharingResponseHeaderFields(SharingEndpoint endpoint, JsonObject body, int clientSiteId) {
+        assertEquals(this.getIdentityScope().toString(), body.getString("identity_scope"));
+        assertNotNull(body.getInteger("allow_clock_skew_seconds"));
```
sunnywu, line 4247: do we wanna check the value against an expected value?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
jon8787: +1, will be good to have a few assertions that unwanted fields are not there

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
mcollins-ttd: There's no change in behaviour, the old version only includes `site_data` if `keySharingEndpointProvideSiteDomainNames && clientSideTokenGenerate`.

src/main/java/com/uid2/operator/Const.java
mcollins-ttd: Most of the consts in this file are only used once. I appreciate your point and don't have a strong opinion on it, but as this seems to be the existing pattern I'd prefer to stick to it. If we want to change the pattern, I'm happy to do that in a separate PR.

jon8787: this is not the existing pattern though :) there's a mix throughout this codebase

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: Maybe not now that we're adding a buffer as well :/

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
mcollins-ttd: That was the old comment just moved to a new spot. I've updated the comment and removed the word "enabled".

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: Done

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: Done

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: Done

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -508,103 +520,143 @@ private String getSharingTokenExpirySeconds() {
...
-                    if(s == null || s.getDomainNames().isEmpty()) {
-                        continue;
-                    }
-                    JsonObject siteObj = new JsonObject();
-                    siteObj.put("id", siteId);
-                    siteObj.put("domain_names", s.getDomainNames().stream().sorted().collect(Collectors.toList()));
-                    sites.add(siteObj);
-                }
-                /*
-                The end result will look something like this:
```
sunnywu, line 581: can we keep this comment in com.uid2.operator.vertx.UIDOperatorVerticle#addSites?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -413,17 +417,41 @@ private void checkEncryptionKeysSharing(JsonObject response, int callersSiteId,
...
+                    }
+                    break;
+                case BIDSTREAM:
+                    assertNull(actualKeysetId);
+                    break;
             }
         }
     }
 
+    private enum SharingEndpoint {
```
sunnywu, line 440: nit: could we call this SdkDownloadEndpoint or something non-Sharing? Sharing is just confusing with SharingClient ...

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -4170,17 +4233,46 @@ void keySharingRotatingKeysets_IDREADER(String testRun, Vertx vertx, VertxTestCo
...
+            final JsonObject body = respJson.getJsonObject("body");
+
+            checkSharingResponseHeaderFields(endpoint, body, clientSiteId);
+
+            checkEncryptionKeys(respJson, endpoint, clientSiteId, expectedKeys.toArray(new KeysetKey[0]));
             testContext.completeNow();
         });
     }
 
+    private void checkSharingResponseHeaderFields(SharingEndpoint endpoint, JsonObject body, int clientSiteId) {
```
sunnywu, line 4248: same comment to renaming Sharing to something else e.g. SdkDownloadResponse

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -4066,21 +4117,33 @@ void keySharingKeysets_CorrectIDS(String testRun, Vertx vertx, VertxTestContext
...
+    @MethodSource("keySharingRotatingKeysets_IDREADER_source")
         // "KeysetAccess"
         //   ID_READER has access to a keyset that has the same site_id as ID_READER's  - direct access
         //   ID_READER has access to a keyset with a missing allowed_sites              - access through sharing
         //   ID_READER has access to a keyset with allowed_sites that includes us       - access through sharing
         //   ID_READER has no access to a keyset that is disabled                       - direct reject
         //   ID_READER has no access to a keyset with an empty allowed_sites            - reject by sharing
         //   ID_READER has no access to a keyset with an allowed_sites for other sites  - reject by sharing
-    void keySharingRotatingKeysets_IDREADER(String testRun, Vertx vertx, VertxTestContext testContext) {
+    void keySharingRotatingKeysets_IDREADER(String testRun, SharingEndpoint endpoint, Vertx vertx, VertxTestContext testContext) {
```
sunnywu, line 4146: why would we test calling /key/sharing endpoint as an ID_READER role? is this intentional? For pretending to be a legacy (c#) UID2Client?

if so, maybe worth commenting here that this tests verify the content downloaded as a Bidstream OR SharingClient for a client with ID READER role, otherwise i just assume this is testing BidstreamClient download content judging by ID READER method name?

src/main/java/com/uid2/operator/Const.java
mcollins-ttd: Discussed offline, we'll keep the constants in this PR.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -106,6 +105,9 @@ public class UIDOperatorVerticle extends AbstractVerticle {
     public final static int MASTER_KEYSET_ID_FOR_SDKS = 9999999; //this is because SDKs have an issue where they assume keyset ids are always positive; that will be fixed.
     public final static long OPT_OUT_CHECK_CUTOFF_DATE = Instant.parse("2023-09-01T00:00:00.00Z").getEpochSecond();
 
+    private final int maxBidstreamLifetimeSeconds;
+    private final int allowClockSkewSeconds;
+    protected int maxSharingLifetimeSeconds;
```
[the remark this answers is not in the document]
mcollins-ttd: I described why we add the buffer in the commit messages that add the override config values.

I'll add a comment here that we should add a buffer due to issues with the current ad token V2 implementation.

I don't want to add a buffer to the fallback values in code to work around the current implementation issue with token lifetimes. It makes the code and tests more complicated. I'd rather add the buffer in config as it's much simpler to test, and it seems like the perfect use for the overrides that we'd already accounted for in the design.

I can't confirm the token lifetime issue is present with v4 tokens, so once we move to v4 the issue may be fixed.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: `KeyEndpoint`? Since they both start `/key/`?

sunnywu: `KeyDownloadEndpoint` ? Key itself is too vague

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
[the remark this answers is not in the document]
mcollins-ttd: As discussed offline, we're making the buffer permanent. I've reverted the config changes and added the buffer in the code.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -144,6 +146,14 @@ public UIDOperatorVerticle(JsonObject config,
         this.keySharingEndpointProvideSiteDomainNames = config.getBoolean("key_sharing_endpoint_provide_site_domain_names", false);
         this._statsCollectorQueue = statsCollectorQueue;
         this.clientKeyProvider = clientKeyProvider;
+        final Integer identityTokenExpiresAfterSeconds = config.getInteger(UIDOperatorService.IDENTITY_TOKEN_EXPIRES_AFTER_SECONDS);
```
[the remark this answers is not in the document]
mcollins-ttd: Added the buffer to the code instead of the config.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
mcollins-ttd: Done

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: Changed to `KeyDownloadEndpoint`

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: Changed the name and added a comment, please let me know if it's still unclear

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -153,6 +155,9 @@ private void setupConfig(JsonObject config) {
         config.put("client_side_token_generate", true);
         config.put("key_sharing_endpoint_provide_site_domain_names", true);
         config.put("client_side_token_generate_log_invalid_http_origins", true);
+
+        config.put(Const.Config.AllowClockSkewSecondsProp, 3600);
+        config.put(Const.Config.MaxBidstreamLifetimeSecondsProp, identityExpiresAfter.toSeconds() + 10);
```
sunnywu, line 160: do you still need to manually add the 10 seconds now that addBidstreamHeaderFields method will add the tolerence already?

mcollins-ttd: I'll update this, it's definitely confusing. _This_ +10 seconds came before the buffer and was intended to be "some custom value".

sunnywu: probably good to add a comment to say the purpose of this custom value

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -68,11 +68,15 @@
 
 import static com.uid2.operator.IdentityConst.*;
 import static com.uid2.operator.service.ResponseUtil.*;
-import static com.uid2.shared.middleware.AuthMiddleware.API_CLIENT_PROP;
 
 public class UIDOperatorVerticle extends AbstractVerticle {
     private static final Logger LOGGER = LoggerFactory.getLogger(UIDOperatorVerticle.class);
     public static final long MAX_REQUEST_BODY_SIZE = 1 << 20; // 1MB
+    /**
+     * There is currently an issue with v2 tokens (and possibly also other ad token versions) where the token lifetime
```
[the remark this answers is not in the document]
mcollins-ttd: Yes, we didn't fix the issue

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
mcollins-ttd: Added `keyBidstreamReturnsCustomMaxBidstreamLifetimeHeader` to test custom config value.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
jon8787: when is keySharingEndpointProvideSiteDomainNames ever false? do we still need that config

jon8787 (approved):

sunnywu (approved):

PR #404, opened 2024-03-07
Remove optout url from private operator

thomasm-ttd:


conf/local-e2e-docker-private-config.json
```diff
@@ -23,7 +23,7 @@
   "validate_service_links": false,
   "optout_s3_bucket": "test-optout-bucket",
   "optout_s3_folder": "optout-v2/",
-  "optout_metadata_path": "http://optout:8081/optout/refresh",
+  "optout_metadata_path": "/optout/refresh",
   "optout_api_uri": "http://optout:8081/optout/replicate",
```
gmsdelmundo, line 27: Do we want this removed as well? Or just the /refresh endpoint is enough?

thomasm-ttd: It's only the refresh endpoint. The other one is more complicated and only used by our operator, not the private ones

gmsdelmundo (approved):

PR #405, opened 2024-03-12
Added support for AWS EUID E2Es

gmsdelmundo:


scripts/aws/EUID_CloudFormation.template.yml
```diff
@@ -1,5 +1,5 @@
 AWSTemplateFormatVersion: 2010-09-09
-Description: EUID CloudFormation template
+Description: EUID 2.0 CloudFormation template
```
thomasm-ttd, line 2: We don't have a version number on EUID. This will create confusion. Please remove

PR #453, opened 2024-03-25
Separate UID2 and EUID EIF build steps

cYKatherine:
Successful run: https://github.com/IABTechLab/uid2-operator/actions/runs/8428783502

### DIFF
--- a/.github/workflows/publish-aws-nitro-enclave-docker.yaml
+++ b/.github/workflows/publish-aws-nitro-enclave-docker.yaml
@@ -140,6 +140,10 @@ jobs:
       - name: Restore timestamps
         uses: thetradedesk/git-restore-mtime-action@v1.3
 
+      - name: Free up space - delete preinstalled tools
+        run: |
+          rm -rf /opt/hostedtoolcache
+
       - name: Build UID2 AWS EIF
         id: build_uid2_eif
         uses: IABTechLab/uid2-operator/.github/actions/build_aws_eif@main
@@ -179,6 +183,10 @@ jobs:
       - name: Restore timestamps
         uses: thetradedesk/git-restore-mtime-action@v1.3
 
+      - name: Free up space - delete preinstalled tools
+        run: |
+          rm -rf /opt/hostedtoolcache
+      
       - name: Build EUID AWS EIF
         id: build_euid_eif
         uses: IABTechLab/uid2-operator/.github/actions/build_aws_eif@main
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-aws-nitro-enclave-docker.yaml
```diff
@@ -112,57 +112,134 @@ jobs:
...
+        uses: thetradedesk/git-restore-mtime-action@v1.3
+
+      - name: Free up space - delete preinstalled tools
+        run: |
+          rm -rf /opt/hostedtoolcache
 
       - name: Build UID2 AWS EIF
         id: build_uid2_eif
-        uses: IABTechLab/uid2-operator/.github/actions/build_aws_eif@v2
+        uses: IABTechLab/uid2-operator/.github/actions/build_aws_eif@main
```
cody-constine-ttd, line 149: We should probably stay with a specific version here.

.github/workflows/publish-aws-nitro-enclave-docker.yaml
```diff
@@ -112,57 +112,134 @@ jobs:
...
         with:
-          name: aws-uid2-deployment-files-${{ steps.version.outputs.new_version }}
+          name: aws-uid2-deployment-files-${{ needs.start.outputs.new_version }}
           path: ${{ env.ARTIFACTS_BASE_OUTPUT_DIR }}/uid2
           if-no-files-found: error
-  
+    outputs:
+      uid2_enclave_id: ${{ steps.build_uid2_eif.outputs.enclave_id }}
+
+  buildEUIDEIF:
```
cody-constine-ttd, line 163: Could we combine this with the above and either just produce both or have an option in a drop down to pick

cYKatherine: I think the idea originally was to always generate an EIF for both UID2 and EUID.

.github/workflows/publish-aws-nitro-enclave-docker.yaml
cYKatherine: `v2` was wrong. For operator we don't have versioning, we have to specify `main` branch. 

Alternatively, we can use `latest`, but the downside is that if we made some changes and didn't do a release, we won't pick up that change for the pipeline. That's why I kept it as `main`.

thomasm-ttd: main is the correct reference to use I think

thomasm-ttd (approved):

PR #481, opened 2024-04-10
created a version that would include take the app name from encrypted payload in CSTGrequest for app name check

sunnywu:


sunnywu:
no longer needed

PR #502, opened 2024-04-18
Install and configure syslog-ng on both the host and enclave

thomasm-ttd:
Using syslog-ng allows us to remove the loki configuration.
Logs are echoed from the enclave to /var/log/messages by default, but the host syslog-ng can be configured to send the logs to other destinations.

Tested by building a new eif and AMI. AMI started using the cloud formation, and it starts correctly. The logs on the ec2 host have the logs from the enclave in them.


scripts/aws/entrypoint.sh
```diff
@@ -91,12 +95,6 @@ fi
 
 cat "${FINAL_CONFIG}"
 
-# -- setup loki
```
cYKatherine, line 94: Why are we removing loki setup?

cYKatherine: Aw okay it was called below. Do we want `loki_enabled` default as true now?

scripts/aws/entrypoint.sh
```diff
@@ -110,6 +108,7 @@ java \
   -Djava.security.egd=file:/dev/./urandom \
   -Djava.library.path=/app/lib \
   -Dvertx-config-path="${FINAL_CONFIG}" \
-  ${SETUP_LOKI_LINE} \
+  -Dvertx.logger-delegate-factory-class-name=io.vertx.core.logging.SLF4JLogDelegateFactory \
+  -Dlogback.configurationFile=./conf/logback.xml" \
   -Dhttp_proxy=socks5://127.0.0.1:3305 \
-  -jar /app/"${JAR_NAME}"-"${JAR_VERSION}".jar
+  -jar /app/"${JAR_NAME}"-"${JAR_VERSION}".jar > /home/start.txt 2>&1
```
cYKatherine, line 114: Why do we want to printout the error? (if I understand `2>&1` correctly)

scripts/aws/syslog-ng/syslog-ng-server.conf
```diff
@@ -0,0 +1,34 @@
+@version: 4.6
```
cYKatherine, line 1: Is there a doc about how we config syslog? If we can add the documentation as a comment it might be useful.

scripts/aws/entrypoint.sh
thomasm-ttd: That switch is being removed completely - the logs are always on and echoed to the ec2 instance

scripts/aws/entrypoint.sh
thomasm-ttd: This captures the startup log entries before logback has started. Sometimes there are error message here that we need before the JVM has actually started.

scripts/aws/syslog-ng/syslog-ng-server.conf
thomasm-ttd: Good point - added a Readme file with links to the docs

cody-constine-ttd (approved):

scripts/aws/Dockerfile
```diff
@@ -14,9 +14,15 @@ ENV IDENTITY_SCOPE=${IDENTITY_SCOPE}
...
+COPY ./syslog-ng-ose-pub.asc /app/dep/
+
 RUN apt update -y \
-    && apt install -y pkg-config libssl-dev net-tools curl jq netcat python3 python3-pip \
-    && rm -rf /var/lib/apt/lists/*
+    && apt install -y pkg-config libssl-dev net-tools curl jq netcat python3 python3-pip libcap2 libivykis0 libjson-c5 libnet1 libwrap0 \
+    && apt-key add /app/dep/syslog-ng-ose-pub.asc \
+    && apt-get install /app/dep/syslog-ng-core_4.6.0-1_amd64.deb \
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-key del 6694369F
```
cYKatherine, line 25: Where did we get this key value from?

scripts/aws/syslog-ng/client/syslog-ng-ose-pub.asc
```diff
@@ -0,0 +1,53 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
```
cYKatherine, line 1: Does this pass the trivy secret scanning?

scripts/aws/Dockerfile
thomasm-ttd: It is the key of the public key we add, so it comes from the committed file

scripts/aws/syslog-ng/client/syslog-ng-ose-pub.asc
thomasm-ttd: yes - was worried it wouldn't, but passed my local testing. It is a public key

cYKatherine (approved):

PR #540, opened 2024-05-06
Add a delay container start

thomasm-ttd:
Delaying the container start until after the sidecar allows the Java application to access the key vault.

### DIFF
--- a/scripts/azure-cc/entrypoint.sh
+++ b/scripts/azure-cc/entrypoint.sh
@@ -2,6 +2,22 @@
 #
 # This script must be compatible with Ash (provided in eclipse-temurin Docker image) and Bash
 
+function wait_for_sidecar() {
+  url="http://169.254.169.254/ping"
+  delay=1
+
+  while true; do
+    if curl -s --connect-timeout 5 "$url" > /dev/null; then
+      echo "side car started"
+      break
+    else
+      echo "side car not started. Retrying in $delay seconds..."
+      sleep $delay
+      delay=$((delay + 1))
+    fi
+  done
+}
+
 TMP_FINAL_CONFIG="/tmp/final-config.tmp"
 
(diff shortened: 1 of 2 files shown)


scripts/azure-cc/entrypoint.sh
```diff
@@ -2,6 +2,22 @@
 #
 # This script must be compatible with Ash (provided in eclipse-temurin Docker image) and Bash
 
+function wait_for_sidecar() {
+  url="http://169.254.169.254/ping"
+  delay=1
+
+  while true; do
```
atarassov-ttd, line 9: Should there be a limit for how long we'd wait for?

thomasm-ttd: If that doesn't return, we can't talk to the key vault, so as soon as the Java apps starts, its going to fail. However, we do have some retries on that, and when that fails, the container exists and will be restarted, so probably a good idea. Will add it

scripts/azure-cc/entrypoint.sh
```diff
@@ -2,6 +2,27 @@
...
+  delay=1
+  max_retries=15
+
+  while true; do
+    if wget -q --spider --tries=1 --timeout 5 "$url" > /dev/null; then
+      echo "side car started"
+      break
+    else
+      echo "side car not started. Retrying in $delay seconds..."
+      sleep $delay
```
atarassov-ttd, line 16: Is the increasing delay intentional here? As a nit, max retries implies we would retry 15 times (16 total attempts), but the logic is to retry 14 times (15 total attempts).

thomasm-ttd: The increasing delay is intentional - the Azure team suggested a backoff strategy for the delay. Fixed the retry count

atarassov-ttd (approved):

PR #546, opened 2024-05-07
Add method OptOutStoreSnapshot.getAdIdOptOutTimestamp

mcollins-ttd:
This method returns the earliest optout timestamp for a given advertising ID.

One commit is a workaround for an issue where calling `toPartition` on an empty heap throws an assertion error. This issue has laid dormant because we didn't have any tests covering `OptOutStoreSnapshot` until now. The workaround is to use `null` instead, so we need some extra checks.

Included are some small improvements:
- Use `sum` instead of `reduce` for clarity.
- Change type of `OptOutStoreSnapshot.fsLocal` from `ICloudStorage` to `DownloadCloudStorage` for easier testing.
- Pass a `Clock` to `OptOutStoreSnapshot` for easier testing.

### DIFF
--- a/conf/default-config.json
+++ b/conf/default-config.json
@@ -17,6 +17,8 @@
   "optout_partition_interval": 86400,
   "optout_max_partitions": 30,
   "optout_heap_default_capacity": 8192,
+  "optout_status_api_enabled": false,
+  "optout_status_max_request_size": 5000,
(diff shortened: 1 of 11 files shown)


asloobq (approved):
Reviewed Matt's changes

PR #557, opened 2024-05-11
[UID2-2005] Add new endpoint to check opt out status by raw UIDs

asloobq:
Manual Testing:
- Tested by starting up server locally and check for validation errors, unauthorized, etc. Couldn't test for positive case as reading deltas from local config is not straight forward.
- Tested in E2E environment by starting all services and running public operator from this branch. Tested positive and negative use cases

Automated Testing
- Added unit tests for both positive and negative cases


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -168,6 +173,8 @@ public UIDOperatorVerticle(JsonObject config,
         this.allowClockSkewSeconds = config.getInteger(Const.Config.AllowClockSkewSecondsProp, 1800);
         this.maxSharingLifetimeSeconds = config.getInteger(Const.Config.MaxSharingLifetimeProp, config.getInteger(Const.Config.SharingTokenExpiryProp));
         this.saltRetrievalResponseHandler = saltRetrievalResponseHandler;
+        this.optOutStatusApiEnabled = config.getBoolean(Const.Config.OptOutStatusApiEnabled, false);
+        this.optOutStatusMaxRequestSize = config.getInteger(Const.Config.OptOutStatusMaxRequestSize, 1000);
```
sunnywu, line 177: the Optout API confluence page stated 5000? any reason why default stays at 1000?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -2115,6 +2115,102 @@ void identityMapBatchRequestTooLarge(String apiVersion, Vertx vertx, VertxTestCo
...
+        return Stream.of(
+            Arguments.arguments(optedOutIdsCase1, 2),
+            Arguments.arguments(optedOutIdsCase2, 0)
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("optOutStatusRequestData")
+    void optOutStatusRequest(Map<String, Long> optedOutIds, int optedOutCount, Vertx vertx, VertxTestContext testContext) {
+        fakeAuth(126, Role.MAPPER);
```
sunnywu, line 2142: could you parameterise the role to test with id_reader/sharer too?

src/test/java/com/uid2/operator/store/OptOutStoreSnapshotTest.java
```diff
@@ -144,8 +166,9 @@ private byte[] entriesToByteArray(List<OptOutEntry> entries) {
             return bytes;
         }
 
-        private JsonObject make1mOptOutEntryConfig() {
+        private JsonObject make1mOptOutEntryConfig(boolean isEnabled) {
```
sunnywu, line 169: maybe rename to `OptOutStatusApiEnabled` as isEnabled is not clear on what is enabled, optout storage/optout lookup/or optout api?

src/test/java/com/uid2/operator/store/OptOutStoreSnapshotTest.java
```diff
@@ -114,6 +114,28 @@ void emptySnapshotUpdatedWithPartitionFilesReturnsCorrectTimestamps(int partitio
             }
         }
 
+        @Test
+        void optoutStatusApiDisabled()  throws CloudStorageException, IOException {
```
sunnywu, line 118: should this also check if the optout/status endpoint is disabled as well?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -2115,6 +2115,102 @@ void identityMapBatchRequestTooLarge(String apiVersion, Vertx vertx, VertxTestCo
...
+                assertEquals(optedOutIds.get(optOutObject.getString("advertising_id")),
+                        optOutObject.getLong("opted_out_since"));
+            }
+            testContext.completeNow();
+        });
+    }
+
+    private static Stream<Arguments> optOutStatusValidationErrorData() {
+        // Test case 1
+        int requestLimit = 1000;
```
mcollins-ttd, line 2169: At the moment if someone changes the default value in the verticle, they will have to run the tests in order to find that they need to change the value here as well.

Can we make this a field of the test class and use it to configure the verticle, so that changes to the default value in the verticle won't cause this test to break?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1678,6 +1689,74 @@ private void recordIdentityMapStatsForServiceLinks(RoutingContext rc, String api
...
+                    optedOutJsonArray.add(optOutJsonObj);
+                }
+            }
+            // Create response and return
+            final JsonObject bodyJsonObj = new JsonObject();
+            bodyJsonObj.put("opted_out", optedOutJsonArray);
+            ResponseUtil.SuccessV2(rc, bodyJsonObj);
+            recordOptOutStatusEndpointStats(rc, rawUID2sInput.size(), optedOutJsonArray.size());
+        } catch (Exception e) {
+            ResponseUtil.Error(ResponseStatus.UnknownError, 500, rc,
```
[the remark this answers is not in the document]
asloobq: hmm...we probably don't. I think it serves as a safe guard for the future if any of the called method implementation changes. So I would like to keep it

src/test/java/com/uid2/operator/store/OptOutStoreSnapshotTest.java
sunnywu: ok sure

sunnywu (approved):

mcollins-ttd (approved):

PR #563, opened 2024-05-15
UID2-3315 Update log names with operator prefix

cYKatherine:


scripts/aws/logrotate/logrotate
thomasm-ttd, line 1: Is there a reason this is not called logrotate.sh?

thomasm-ttd (approved):

PR #585, opened 2024-05-26
Add metrics info for Mobile CSTG

caroline-ttd:
```
uid2_token_response_status_count_total{advertising_token_version="V2", application="uid2-operator", cstg="true", instance="host.docker.internal:9080", job="services", platformType="HasOriginHeader", site_id="123", site_name="MegaTest Site", token_endpoint="ClientSideTokenGenerateV2", token_response_status="Success"}

uid2_token_response_status_count_total{advertising_token_version="null", application="uid2-operator", cstg="true", instance="host.docker.internal:9080", job="services", platformType="InApp", site_id="123", site_name="MegaTest Site", token_endpoint="ClientSideTokenGenerateV2", token_response_status="InvalidAppName"}
```

Slack discussions are here: https://thetradedesk.slack.com/archives/C051ELQ8JAC/p1716760305149009

Add the platformType for `/token/client-generate`,` /token/refresh` and `/token/generate`

### DIFF
--- a/src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
+++ b/src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
@@ -43,8 +43,8 @@ public class TokenResponseStatsCollector {
     }
 
     public enum PlatformType {
-        InApp, // Request containing the "X-UID2-Client-Version" header, typically originating from Android, iOS, or tvOS (Apple TV).
-        HasOriginHeader, // Request containing the "original" header, originating from the web.
+        InApp, // Request has the "X-UID2-Client-Version" header, which contains "Android", "ios" or "tvos", typically originating from Android, iOS, or tvOS (Apple TV).
+        HasOriginHeader, // Request has the "origin" header, originating from the web.
         Other // Everything else, such as requests originating from the server side.
     }
 
--- a/src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
+++ b/src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
@@ -42,11 +42,17 @@ public class TokenResponseStatsCollector {
         Unauthorized
     }
 
-    public static void record(ISiteStore siteStore, Integer siteId, Endpoint endpoint, TokenVersion advertisingTokenVersion, ResponseStatus responseStatus) {
-        recordInternal(siteStore, siteId, endpoint, responseStatus, advertisingTokenVersion, endpoint == Endpoint.ClientSideTokenGenerateV2);
+    public enum PlatformType {
+        Unknown,
+        Mobile,
+        Web
     }
 
-    private static void recordInternal(ISiteStore siteStore, Integer siteId, Endpoint endpoint, ResponseStatus responseStatus, TokenVersion advertisingTokenVersion, boolean isCstg) {
+    public static void record(ISiteStore siteStore, Integer siteId, Endpoint endpoint, TokenVersion advertisingTokenVersion, ResponseStatus responseStatus, PlatformType platformType) {
+        recordInternal(siteStore, siteId, endpoint, responseStatus, advertisingTokenVersion, endpoint == Endpoint.ClientSideTokenGenerateV2, platformType);
+    }
+
+    private static void recordInternal(ISiteStore siteStore, Integer siteId, Endpoint endpoint, ResponseStatus responseStatus, TokenVersion advertisingTokenVersion, boolean isCstg, PlatformType platformType) {
         if (siteId == null) return;
 
         var builder = Counter
@@ -57,7 +63,8 @@ public class TokenResponseStatsCollector {
                             "token_endpoint", String.valueOf(endpoint),
                             "token_response_status", String.valueOf(responseStatus),
                             "advertising_token_version", String.valueOf(advertisingTokenVersion),
-                            "cstg", isCstg ? "true" : "false");
+                            "cstg", isCstg ? "true" : "false",
+                            "platformType", String.valueOf(platformType));
 
         builder.register(Metrics.globalRegistry).increment();
     }
@@ -65,14 +72,14 @@ public class TokenResponseStatsCollector {
     public static void recordRefresh(ISiteStore siteStore, Integer siteId, Endpoint endpoint, RefreshResponse refreshResponse) {
         if (!refreshResponse.isRefreshed()) {
             if (refreshResponse.isOptOut() || refreshResponse.isDeprecated()) {
-                recordInternal(siteStore, siteId, endpoint, ResponseStatus.OptOut, refreshResponse.getTokens().getAdvertisingTokenVersion(), refreshResponse.isCstg());
+                recordInternal(siteStore, siteId, endpoint, ResponseStatus.OptOut, refreshResponse.getTokens().getAdvertisingTokenVersion(), refreshResponse.isCstg(), PlatformType.Unknown);
             } else if (refreshResponse.isInvalidToken()) {
-                recordInternal(siteStore, siteId, endpoint, ResponseStatus.InvalidToken, refreshResponse.getTokens().getAdvertisingTokenVersion(), refreshResponse.isCstg());
+                recordInternal(siteStore, siteId, endpoint, ResponseStatus.InvalidToken, refreshResponse.getTokens().getAdvertisingTokenVersion(), refreshResponse.isCstg(), PlatformType.Unknown);
             } else if (refreshResponse.isExpired()) {
-                recordInternal(siteStore, siteId, endpoint, ResponseStatus.ExpiredToken, refreshResponse.getTokens().getAdvertisingTokenVersion(), refreshResponse.isCstg());
+                recordInternal(siteStore, siteId, endpoint, ResponseStatus.ExpiredToken, refreshResponse.getTokens().getAdvertisingTokenVersion(), refreshResponse.isCstg(), PlatformType.Unknown);
             }
         } else {
-            recordInternal(siteStore, siteId, endpoint, ResponseStatus.Success, refreshResponse.getTokens().getAdvertisingTokenVersion(), refreshResponse.isCstg());
+            recordInternal(siteStore, siteId, endpoint, ResponseStatus.Success, refreshResponse.getTokens().getAdvertisingTokenVersion(), refreshResponse.isCstg(), PlatformType.Unknown);
         }
     }
 }
--- a/src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
+++ b/src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
@@ -43,9 +43,9 @@ public class TokenResponseStatsCollector {
     }
 
     public enum PlatformType {
-        Unknown,
-        Mobile,
-        Web
+        InApp,
+        HasOriginHeader,
+        Other
     }
 
     public static void record(ISiteStore siteStore, Integer siteId, Endpoint endpoint, TokenVersion advertisingTokenVersion, ResponseStatus responseStatus, PlatformType platformType) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
```diff
@@ -42,11 +42,17 @@ public enum ResponseStatus {
         Unauthorized
     }
 
-    public static void record(ISiteStore siteStore, Integer siteId, Endpoint endpoint, TokenVersion advertisingTokenVersion, ResponseStatus responseStatus) {
-        recordInternal(siteStore, siteId, endpoint, responseStatus, advertisingTokenVersion, endpoint == Endpoint.ClientSideTokenGenerateV2);
+    public enum PlatformType {
+        Unknown,
```
jon8787, line 46: I noticed you use PlatformType.Unknown in a lot of places. Are they all really unknown? 
If browser adds an Origin header to requests such as /token/refresh, then do we know it's Web?

src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
```diff
@@ -42,11 +42,17 @@ public enum ResponseStatus {
...
+        Mobile,
```
sunnywu, line 47: should we rename this App instead? i know we have been mentioning mobile everywhere but strictly speaking it's Andorid/iOS/tvOS/etc native apps. 

We only call it mobile because of our target audience (mobile app publishers) but since this is internal use i think we can call it the right term Apps?

src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
```diff
@@ -42,11 +42,17 @@ public enum ResponseStatus {
...
+        Web
```
sunnywu, line 48: based on @jon8787 's comment above, you might wanna comment here and say Web is defined as request with a http origin header.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1787,13 +1795,24 @@ private RefreshResponse refreshIdentity(RoutingContext rc, String tokenStr) {
...
     public static String getSiteName(ISiteStore siteStore, Integer siteId) {
-        if (siteId == null) return "unknown";
+        if (siteId == null) return "X-UID2-Client-Version";
         if (siteStore == null) return "unknown"; //this is expected if CSTG is not enabled, eg for private operators
 
         final Site site = siteStore.getSite(siteId);
         return (site == null) ? "unknown" : site.getName();
     }
 
+    private TokenResponseStatsCollector.PlatformType getPlatformType(RoutingContext rc) {
```
sunnywu, line 1805: i think this should check app_name first even tho this method is used for other non-cstg endpoints e.g.:

if(rc.request().getHeader("app_name" ) != null)
{
  return InApp;
}
// and then continue rest of the logic below

it is still quicker to check "app_name" header instead of Android/ios string search and even tho it's unnecessary for non-cstg endpoints

src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
```diff
@@ -42,11 +42,17 @@ public enum ResponseStatus {
         Unauthorized
     }
 
-    public static void record(ISiteStore siteStore, Integer siteId, Endpoint endpoint, TokenVersion advertisingTokenVersion, ResponseStatus responseStatus) {
-        recordInternal(siteStore, siteId, endpoint, responseStatus, advertisingTokenVersion, endpoint == Endpoint.ClientSideTokenGenerateV2);
+    public enum PlatformType {
+        InApp,
```
sunnywu, line 46: put some comments on what each value means would be helpful

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1786,13 +1795,24 @@ private RefreshResponse refreshIdentity(RoutingContext rc, String tokenStr) {
     }
 
     public static String getSiteName(ISiteStore siteStore, Integer siteId) {
-        if (siteId == null) return "unknown";
+        if (siteId == null) return "X-UID2-Client-Version";
```
jon8787, line 1798: bit confused about this change - what's the reason to return this header string?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
jon8787: Could the above shortcut be put in `handleClientSideTokenGenerateImpl` instead?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1786,13 +1795,24 @@ private RefreshResponse refreshIdentity(RoutingContext rc, String tokenStr) {
...
         if (siteStore == null) return "unknown"; //this is expected if CSTG is not enabled, eg for private operators
 
         final Site site = siteStore.getSite(siteId);
         return (site == null) ? "unknown" : site.getName();
     }
 
+    private TokenResponseStatsCollector.PlatformType getPlatformType(RoutingContext rc) {
+        final String clientVersion = rc.request().getHeader("X-UID2-Client-Version");
```
jon8787, line 1806: this is a magic string, we should use `ClientVersionHeader`

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1786,13 +1795,24 @@ private RefreshResponse refreshIdentity(RoutingContext rc, String tokenStr) {
...
+        if (clientVersion != null && (clientVersion.contains("Android") || clientVersion.contains("ios") || clientVersion.contains("tvos"))) {
```
jon8787, line 1807: curious, what's `tvos`, if it's a CTV string, are there others?

sunnywu: tvOS is essentially iOS of Apple TV. 
https://en.wikipedia.org/wiki/TvOS
depends on what os the sdk is built with the client version would say ios or tvos.


no there is no other platform string variants that will appear on X-UID2-Client-Version at this stage.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
sunnywu: yes (and my preference) but i thought you suggested to keep it all in 1 function?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
sunnywu: Just a suggestion not advocating hard for it: maybe we should hardcode the android/ios/tvos list somewhere and then use it like:

```
public static boolean containsKnownPlatformType(String str, List<String> substrings) {
  return substrings.stream().anyMatch(substring -> str.contains(substring));
}

```
?

jon8787: We should definitely do that once there's a 2nd use-case. Not sure I'd use `stream()` though.

is "tvos" and "ios" case sensitive?

sunnywu: well we are using String.contains which is case sensitive ?

https://github.com/IABTechLab/uid2-ios-sdk/blob/91c290d29a7093cfc209eca493d1fee80c17e16a/Sources/UID2/UID2Client.swift#L36-L38

jon8787: Maybe we should add this URL as a comment here. All lower-case and `tvos` was surprising to me

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
jon8787: Yes, this is still 1 function. I wrote "in CSTG you can avoid calling it if you want to look for "app_name" first". As in, you can avoid calling `getPlatformType` if you do the "app_name" shortcut first.

caroline-ttd: I added `platformType = request.getAppName() != null ? TokenResponseStatsCollector.PlatformType.InApp : getPlatformType(rc); ` in the function `handleClientSideTokenGenerateImpl`

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: My bad, it is a typo, I changed it back.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: Thank you for the reminder, I updated it to `clientVersionHeader`

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: I added the URL as a comment and created a HashSet for "Android", "ios", "tvOS".

src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
```diff
@@ -42,11 +42,17 @@ public enum ResponseStatus {
         Unauthorized
     }
 
-    public static void record(ISiteStore siteStore, Integer siteId, Endpoint endpoint, TokenVersion advertisingTokenVersion, ResponseStatus responseStatus) {
-        recordInternal(siteStore, siteId, endpoint, responseStatus, advertisingTokenVersion, endpoint == Endpoint.ClientSideTokenGenerateV2);
+    public enum PlatformType {
+        InApp, // Request containing the "X-UID2-Client-Version" header, typically originating from Android, iOS, or tvOS (Apple TV).
```
jon8787, line 46: Not just "X-UID2-Client-Version", since server and JS SDKs contain this as well. It has to specifically contain ios/android/tvos in this header.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
jon8787: Does it need to be a HashSet? are we ever calling `hashSet.contains()` ?

caroline-ttd: Nope, we are only using String.contains `SUPPORTED_IN_APP.stream().anyMatch(clientVersionHeader::contains))`. 

It can also be an array list, I just feel as the values are unique, maybe good to use HashSet?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1793,6 +1806,17 @@ public static String getSiteName(ISiteStore siteStore, Integer siteId) {
         return (site == null) ? "unknown" : site.getName();
     }
 
+    private TokenResponseStatsCollector.PlatformType getPlatformType(RoutingContext rc) {
+        final String clientVersionHeader = rc.request().getHeader("X-UID2-Client-Version");
```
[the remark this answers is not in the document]
caroline-ttd: Oh, sorry, I got it. Will change.

src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
```diff
@@ -42,11 +42,17 @@ public enum ResponseStatus {
...
+        HasOriginHeader, // Request containing the "original" header, originating from the web.
```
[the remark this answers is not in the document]
caroline-ttd: Thank you for catching it!

src/main/java/com/uid2/operator/monitoring/TokenResponseStatsCollector.java
caroline-ttd: Yeah, that makes sense.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: I don't think we need  `hashSet.contains()` function, I will change it to an array.

jon8787: Yes, we're mainly doing iteration, with is faster with an array (not that it matters much). Uniqueness is a good feature but since we're providing all the values, we can already ensure they're unique.

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -518,7 +525,7 @@ private OriginOrAppNameValidationResult validateOriginOrAppName(RoutingContext r
                     : OriginOrAppNameValidationResult.invalidAppName(appName);
         }
 
-        final String origin = rc.request().getHeader("origin");
+        final String origin = rc.request().getHeader(ORIGIN_HEADER);
```
jon8787, line 528: It's actually the whole expression `rc.request().getHeader(ORIGIN_HEADER);` that's duplicated many times, not just the "origin" string. So we should instead have a method `getOriginHeader(rc)`.

But that raises a question - why do we sometimes have `rc.request().headers().contains(ORIGIN_HEADER));` instead of `getHeader`?

jon8787 (approved):

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: `rc.request().headers().contains(ORIGIN_HEADER));` from the [MR](https://github.com/IABTechLab/uid2-operator/commit/63faf62f8735c2754516d551f76fa2913743c4cd#diff-4253d50195965a8f4ceaedaf401049047f4b84a1158cf9575d805153da945b4f). 

Hi Vishal, looks like you made this change, do you have any input @vishalegbert-ttd

PR #591, opened 2024-05-28
Ian UID2-1846 version in query param

Ian-Nara:


src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -5005,4 +5005,39 @@ void secureLinkValidationFailsReturnsIdentityError(Vertx vertx, VertxTestContext
             testContext.completeNow();
         });
     }
+
+    @ParameterizedTest // note that this test will be removed when we switch to logging versions
+    @ValueSource(strings = {"euid-sdk-1.0.0", "openid-sdk-1.0", "uid2-esp-0.0.1a", "uid2-sdk-0.0.1a",
```
jon8787, line 5010: what's the behavior when it's not one of these strings? should we add a test for that too?

Ian-Nara: Yeah sounds good, added test for it

thomasm-ttd (approved):

PR #602, opened 2024-06-07
Add logging to get information about refresh token version

cYKatherine:


src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
```diff
@@ -9,17 +9,21 @@
...
+import java.util.Map;
 
 public class EncryptedTokenEncoder implements ITokenEncoder {
-
     private final KeyManager keyManager;
 
     public EncryptedTokenEncoder(KeyManager keyManager) {
         this.keyManager = keyManager;
     }
+    private final Map<String, Counter> _refreshTokenVersionCounters = new HashMap<>();
```
jon8787, line 26: sure we need a map? some recent changes to Operator removed maps like this one

cYKatherine: This is to make sure we don't create duplicated metrics on the same site id

jon8787: but micrometer already ensures this right?

vishalegbert-ttd: Hey Jonathan, is there an example of this? A lot of our metrics use a dictionary to keep track of how various labels map to different counters

jon8787: Let me try to find. @mcollins-ttd recently removed one of these dictionaries from memory

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
```diff
@@ -259,13 +263,24 @@ public AdvertisingToken decodeAdvertisingTokenV3orV4(Buffer b, byte[] bytes, Tok
         );
     }
 
+    private void recordRefreshTokenVersionCount(String siteId, TokenVersion tokenVersion) {
+        _refreshTokenVersionCounters.computeIfAbsent(siteId, id -> Counter
+                .builder(String.format("uid2_refresh_token_%s_served_count", tokenVersion.toString().toLowerCase()))
```
vishalegbert-ttd, line 268: Now that I think about it.. version should be a label as well to make it easier to visualise the metric across versions. But we also don't update refresh version often so I'll leave it up to you

vishalegbert-ttd: So we'll need to map <version, site ID> to counter instead

vishalegbert-ttd (approved):

src/main/java/com/uid2/operator/service/EncryptedTokenEncoder.java
jon8787: agree, definitely better to use label instead of a dynamic metric name

PR #633, opened 2024-06-16
Respond with 400 instead of 500 when CSTG request validation fails

caroline-ttd:


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -453,6 +453,10 @@ else if(emailHash != null) {
             input = InputUtil.normalizePhoneHash(phoneHash);
         }
 
+        if (this.phoneSupport ? !checkTokenInputV1(input, rc) : !checkTokenInput(input, rc)) {
```
jon8787, line 456: This expression `this.phoneSupport ? !checkTokenInputV1(input, rc) : !checkTokenInput(input, rc)` appears 6 times in this file and ideally should be wrapped in a method

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -4162,6 +4162,56 @@ void cstgNoActiveKey(Vertx vertx, VertxTestContext testContext) throws NoSuchAlg
...
+                testContext,
+                respJson -> {
+                    assertFalse(respJson.containsKey("body"));
+                    assertEquals("Invalid Identifier", respJson.getString("message"));
+                    testContext.completeNow();
+                });
+    }
+
+    @Test
+    void cstgInvalidPhoneHashInput(Vertx vertx, VertxTestContext testContext) throws NoSuchAlgorithmException, InvalidKeyException {
```
jon8787, line 4191: possible to eliminate some of the duplication? the tests are identical other than the input. Can the input be parameterized?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
mcollins-ttd: The only difference between `checkTokenInputV1` and `checkTokenInput` is the error message returned.

Can we remove `checkTokenInputV1` and do the `this.phoneSupport` check in `checkTokenInput`?

caroline-ttd: I removed the `checkTokenInputV1` and did the `this.phoneSuppor`t check in `checkTokenInput`. Also invert the method.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
caroline-ttd: That is a good idea, I used the `@ParameterizedTest`

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -453,6 +453,10 @@ else if(emailHash != null) {
             input = InputUtil.normalizePhoneHash(phoneHash);
         }
 
+        if (checkTokenInput(input, rc)) {
```
jon8787, line 456: `if (checkTokenInput)` makes it seem like it succeeded. How about we rename this method to `checkForInvalidTokenInput`?

jon8787 (changes requested):

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
caroline-ttd: I rename checkTokenInput to checkForInvalidTokenInput. true for invalid input, false for valid input.

jon8787 (approved):

PR #702, opened 2024-07-11
Adding start of the eks work

cody-constine-ttd:


scripts/aws/eks/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM amazonlinux:2023
+RUN yum install aws-nitro-enclaves-cli-devel jq -y
```
thomasm-ttd, line 2: Why do we need the devel tools?

PR #780, opened 2024-07-23
Made eks image smaller

cody-constine-ttd:


scripts/aws/eks/pod/Dockerfile
```diff
@@ -1,7 +1,40 @@
-FROM amazonlinux:2023
+######## full image ########
+FROM amazonlinux:2023 as full_image
+
+RUN yum install -y aws-nitro-enclaves-cli && \
+    yum install aws-nitro-enclaves-cli-devel jq -y
```
cYKatherine, line 5: As @thomasm-ttd mentioned in my last PR `aws-nitro-enclaves-cli-devel` is probably not needed

scripts/aws/eks/pod/Dockerfile
```diff
@@ -1,7 +1,40 @@
-FROM amazonlinux:2023
+######## full image ########
+FROM amazonlinux:2023 as full_image
+
+RUN yum install -y aws-nitro-enclaves-cli && \
```
cYKatherine, line 4: Also let's use `dnf` instead of `yum`: https://docs.aws.amazon.com/linux/al2023/ug/package-management.html

scripts/aws/eks/pod/Dockerfile
cody-constine-ttd: This is amazons recommendation. The later code removes anything we don't need.

cody-constine-ttd: I will see if removing it changes anything.

thomasm-ttd: I would like to remove it unless we absolutely need it - not from a size thing, but it provides more tools and surface area for compromise on the image than we need.

thomasm-ttd (approved):

PR #799, opened 2024-07-24
Ian UID2-3703 retrieve configs from identity service

Ian-Nara:


thomasm-ttd (approved):

scripts/aws/eks/enclave/entrypoint.sh
```diff
@@ -74,13 +52,9 @@ else
...
 fi
 
-get_config_value() {
-  jq -r ".\"$1\"" ${FINAL_CONFIG}
-}
-
 # -- replace base URLs if both CORE_BASE_URL and OPTOUT_BASE_URL are provided
 # -- using hardcoded domains is fine because they should not be changed frequently
-if [ -n "${CORE_BASE_URL}" ] && [ -n "${OPTOUT_BASE_URL}" ] && [ "${DEPLOYMENT_ENVIRONMENT}" != "prod" ]; then
+if [ -n "${CORE_BASE_URL}" ] && [ "${CORE_BASE_URL}" != "null" ] && [ -n "${OPTOUT_BASE_URL}" ] && [ "${OPTOUT_BASE_URL}" != "null" ] && [ "${DEPLOYMENT_ENVIRONMENT}" != "prod" ]; then
```
cYKatherine, line 57: Intereting, why do we need to check null again? Does `-n` not checking null?

scripts/aws/eks/pod/proxies.host.yaml
```diff
@@ -8,12 +8,12 @@ socks5h-proxy:
 operator-service:
   service: direct
   listen: tcp://0.0.0.0:80
-  connect: vsock://42:8080
+  connect: vsock://16:8080
```
cYKatherine, line 11: why are we changing this?

scripts/aws/eks/enclave/entrypoint.sh
Ian-Nara: jq will return "null" if the key is not found

scripts/aws/eks/pod/proxies.host.yaml
Ian-Nara: eks/pod/entrypoint.sh has
CID=16
looks that @thomasm-ttd picked this number

PR #856, opened 2024-08-06
Install syslog-ng in the pod

thomasm-ttd:
Installed the packages for syslog-ng.
Built the EIF and pod
Ran the pod in the test cluster - started and the logs were echoed to the k8s logs


scripts/aws/eks/pod/Dockerfile
```diff
@@ -1,31 +1,43 @@
 FROM amazonlinux:2023
+RUN dnf check-update && dnf update && dnf install nmap-ncat libxcrypt-compat python3 aws-nitro-enclaves-cli logrotate iproute net-tools python3.11 python3.11-pip shadow-utils -y
```
cYKatherine, line 2: Why do we need to install python3.11 here? I think AL2023 includes python3 by default: https://docs.aws.amazon.com/linux/al2023/ug/python.html any reason we want to use 3.11 specifically?

scripts/aws/eks/pod/Dockerfile
```diff
@@ -1,31 +1,43 @@
...
+ARG IVYKIS_RPM="ivykis-0.43-1.amzn2023.x86_64.rpm"
+ARG LIBNET_RPM="libnet-1.2-2.amzn2023.0.2.x86_64.rpm"
+ARG PUBKEY="pubkey.gpg"
+ARG SYSLOG_NG_RPM="syslog-ng-4.7.1.104.gcc5a7d9-1.amzn2023.x86_64.rpm"
+ARG SYSLOG_NG_LOGROTATE_RPM="syslog-ng-logrotate-4.7.1.104.gcc5a7d9-1.amzn2023.x86_64.rpm"
 
 COPY ./sockd /home/
 COPY ./sockd_eks.conf /etc/sockd.conf
 COPY ./vsockpx /home
+COPY ./${IVYKIS_RPM} /home/syslog-ng/
```
cYKatherine, line 13: Have we committed these rpm files?

scripts/aws/eks/pod/entrypoint.sh
```diff
@@ -37,20 +37,48 @@ function setup_dante() {
...
+            echo "Config server still starting..."
+        fi
+        sleep 5
+    done
 }
 
 function run_enclave() {
     echo "starting enclave..."
-    nitro-cli run-enclave --cpu-count $CPU_COUNT --memory $MEMORY_MB --eif-path $EIF_PATH --enclave-cid $CID --enclave-name simple-eif --debug-mode --attach-console
+    nitro-cli run-enclave --cpu-count $CPU_COUNT --memory $MEMORY_MB --eif-path $EIF_PATH --enclave-cid $CID --enclave-name uid2-operator
```
cYKatherine, line 66: Originally we got the cpu and memory resource settings from the secretsmanager, do we want to maintain that mechanism or just hardcode it for our new EKS?

cYKatherine (approved):

scripts/aws/eks/pod/Dockerfile
thomasm-ttd: I only moved the install.

scripts/aws/eks/pod/entrypoint.sh
thomasm-ttd: No, definitely get them from the secret. I will fix this

PR #891, opened 2024-08-12
Fixed 500 from emails being sent as a single email instead of an arra…

cody-constine-ttd:
…y in identity map


thomasm-ttd (commented):
Would it make sense to move the validation code into a shared location (and preferrably not in the verticle as that is getting very long) to prevent the duplication?

cody-constine-ttd:
> Would it make sense to move the validation code into a shared location (and preferrably not in the verticle as that is getting very long) to prevent the duplication?

I purposed a solution. Not entirely sure about having the util return to the client, but not sure there is a better way to do it.

Ian-Nara (approved):

PR #965, opened 2024-09-12
Ian 4054 aws config updates

Ian-Nara:


scripts/aws/config-server/app.py
```diff
@@ -14,6 +15,15 @@ def get_config():
                 secret_value_json["core_base_url"] = secret_value_json["core_base_url"].lower()
             if "optout_base_url" in secret_value_json:
                 secret_value_json["optout_base_url"] = secret_value_json["optout_base_url"].lower()
+            if "operator_type" in secret_value_json and secret_value_json['operator_type'] == "public":
```
thomasm-ttd, line 18: This compare should not be case sensitive

PR #996, opened 2024-09-18
Cbc UI d2 4134 entrypoint stops when enclave stops

cody-constine-ttd:


scripts/aws/eks-pod/entrypoint.sh
```diff
@@ -111,4 +111,9 @@ wait_for_config
 update_config
 run_enclave
 
-sleep infinity
\ No newline at end of file
+ENCLAVE_ID=$(nitro-cli describe-enclaves | jq -r ".[0].EnclaveID")
```
Ian-Nara, line 114: could we put a sleep for like 60s before the first check. Have had some crashes from it not showing up fast enough

caroline-ttd (approved):

PR #1039, opened 2024-09-26
Refactor enclave entrypoint

thomasm-ttd:
Refactored the entrypoint to remove the duplicate call to getConfig
Tested in Test and operator starts correctly. Checked the logs, and the configuration is not echoed to the logs, as expected.

### DIFF
--- a/scripts/aws/entrypoint.sh
+++ b/scripts/aws/entrypoint.sh
@@ -29,9 +29,16 @@ echo "Loading config from identity service via proxy..."
 #wait for config service, then download config
 OVERRIDES_CONFIG="/app/conf/config-overrides.json"
 
+RETRY_COUNT=0
+MAX_RETRY=20
 until curl -s -f -o "${OVERRIDES_CONFIG}" -x socks5h://127.0.0.1:3305 http://127.0.0.1:27015/getConfig
 do
   echo "Waiting for config service to be available"
+  RETRY_COUNT=$(( RETRY_COUNT + 1))
+  if [ $RETRY_COUNT -gt $MAX_RETRY ]; then
+      echo "Config Server did not return a response. Exiting"
+      exit 1
+  fi
   sleep 2
 done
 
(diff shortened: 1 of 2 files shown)


scripts/aws/entrypoint.sh
```diff
@@ -22,38 +25,23 @@ echo "Starting syslog-ng..."
...
-  if [ $RETRY_COUNT -gt $MAX_RETRY ]; then
-      echo "Config Server did not return a response. Exiting"
-      exit 1
-  fi
-  sleep 5
 
+#wait for config service, then download config
+export OVERRIDES_CONFIG="/app/conf/config-overrides.json"
+
+until curl -s -f -o "${OVERRIDES_CONFIG}" -x socks5h://127.0.0.1:3305 http://127.0.0.1:27015/getConfig
```
cYKatherine, line 32: Just checking have we made sure the `api_token` is not logged?

thomasm-ttd: Yes, I was specifically looking for it, and it is not logged. We now just write the file directly

cYKatherine (approved):

scripts/aws/entrypoint.sh
atarassov-ttd: Should there be a limit for the number of retries?

scripts/aws/entrypoint.sh
```diff
@@ -22,38 +25,23 @@ echo "Starting syslog-ng..."
...
-  fi
-  RETRY_COUNT=$(( RETRY_COUNT + 1))
-  if [ $RETRY_COUNT -gt $MAX_RETRY ]; then
-      echo "Config Server did not return a response. Exiting"
-      exit 1
-  fi
-  sleep 5
 
+#wait for config service, then download config
+export OVERRIDES_CONFIG="/app/conf/config-overrides.json"
```
atarassov-ttd, line 30: Is export required here? Do any of the scripts/apps called below rely on this variable being set or is it used only within this script?

thomasm-ttd: Removed the export

PR #1097, opened 2024-10-16
1. 

sunnywu:
0. Read changes [here ](https://github.com/IABTechLab/uid2-operator/pull/1081/files)first
1. Removed privacyBits and establishedAt from UserIdentity/HashedDiidentity/RawUidIdentity but only kept inside FirstLevelHashIdentity - these fields are only relevant when a first level hash is generated for token generation and then passed along into ad token generation logic.

2. As a result,  privacyBits and establishedAt are generated from either a. in a brand new token generation call (by default should be 1 and the timestamp at the time of the call) b. or, during token refresh logic, these fields will be inherited from previous refresh token.
3. Updated a lot of codes as a result (more refinement required such as not hardcoding the "1, Instant.now()" in a lot of places)

### DIFF
--- a/src/main/java/com/uid2/operator/model/IdentityRequest.java
+++ b/src/main/java/com/uid2/operator/model/IdentityRequest.java
@@ -2,19 +2,28 @@ package com.uid2.operator.model;
 
 import com.uid2.operator.model.userIdentity.HashedDiiIdentity;
 
+import java.time.Instant;
+
 public final class IdentityRequest {
     public final SourcePublisher sourcePublisher;
     public final HashedDiiIdentity hashedDiiIdentity;
     public final OptoutCheckPolicy optoutCheckPolicy;
 
+    public final int privacyBits;
+    public final Instant establishedAt;
+
     public IdentityRequest(
             SourcePublisher sourcePublisher,
             HashedDiiIdentity hashedDiiIdentity,
-            OptoutCheckPolicy tokenGeneratePolicy)
+            OptoutCheckPolicy tokenGeneratePolicy,
+            int privacyBits,
+            Instant establishedAt)
     {
         this.sourcePublisher = sourcePublisher;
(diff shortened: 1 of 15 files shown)


src/main/java/com/uid2/operator/model/IdentityRequest.java
```diff
@@ -26,6 +27,18 @@ public IdentityRequest(
...
+    public IdentityRequest(
+            SourcePublisher sourcePublisher,
+            HashedDiiIdentity hashedDiiIdentity,
+            OptoutCheckPolicy tokenGeneratePolicy)
+    {
+        this.sourcePublisher = sourcePublisher;
+        this.hashedDiiIdentity = hashedDiiIdentity;
+        this.optoutCheckPolicy = tokenGeneratePolicy;
+        this.privacyBits = PrivacyBits.DEFAULT_PRIVACY_BIT_VALUE;
+        this.establishedAt = Instant.now();
```
mcollins-ttd, line 39:
```suggestion
        this(sourcePublisher, hashedDiiIdentity, tokenGeneratePolicy, PrivacyBits.DEFAULT_PRIVACY_BIT_VALUE, Instant.now());

```

src/main/java/com/uid2/operator/util/PrivacyBits.java
```diff
@@ -3,6 +3,9 @@
 
 public class PrivacyBits {
 
+    // For historical reason this bit is set
+    public static final int DEFAULT_PRIVACY_BIT_VALUE = 1;
```
mcollins-ttd, line 7: Maybe this should be `PrivacyBits` instead?
```java
    public static final PrivacyBits DEFAULT = PrivacyBits.fromInt(1);
```

sunnywu: removed all uses of int PrivacyBits and use PrivacyBits class instead. Added unit tests to test the class

PR #1116, opened 2024-10-26
Allow optout bootstrapping to happen in 2 iterations

asloobq:


asloobq:
Decided not to do it this way

PR #1145, opened 2024-11-07
find participants on old sdks

Ian-Nara:


src/main/java/com/uid2/operator/vertx/ClientVersionCapturingHandler.java
```diff
@@ -49,4 +68,22 @@ private static String getFileNameWithoutExtension(Path path) {
         final String fileName = path.getFileName().toString();
         return fileName.indexOf(".") > 0 ? fileName.substring(0, fileName.lastIndexOf(".")) : fileName;
     }
+
+    private static String extractBearerToken(final String headerValue) {
```
cody-constine-ttd, line 72: Isn't his built into vertx?

Ian-Nara: We implement it ourself :( RoutingContext doesn't provide for this.

cody-constine-ttd (approved):

PR #1159, opened 2024-11-18
Add validations on EC2 enclave pre-init

abuabraham-ttd:
UID2- 4364


scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+from botocore.exceptions import ClientError
+import sys
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+from confidential_compute import ConfidentialCompute
+
+class EC2(ConfidentialCompute):
+
+    def __init__(self):
+        super().__init__()
+        self.config = {}
```
atarassov-ttd, line 21: Below the member variable being used appears to be `self.configs`.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def __get_aws_token(self):
+        try:
+            token_url = "http://169.254.169.254/latest/api/token"
+            token_response = requests.put(token_url, headers={"X-aws-ec2-metadata-token-ttl-seconds": "3600"}, timeout=2)
+            return token_response.text
+        except Exception as e:
+            return "blank"
```
atarassov-ttd, line 29: Why not let the exception propagate through? Why is it valid to swallow the error here?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+    def __get_current_region(self):
+        token = self.__get_aws_token()
+        metadata_url = "http://169.254.169.254/latest/dynamic/instance-identity/document"
+        headers = {"X-aws-ec2-metadata-token": token}
+        try:
+            response = requests.get(metadata_url, headers=headers,timeout=2)
+            if response.status_code == 200:
+                return response.json().get("region")
+            else:
+                print(f"Failed to fetch region, status code: {response.status_code}")
```
atarassov-ttd, line 40: Here and in other places below, why are errors being swallowed and script execution allowed to proceed further?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+        except Exception as e:
+            raise Exception(f"Region not found, are you running in EC2 environment. {e}")
+
+    def _get_secret(self, secret_identifier):
+        client = boto3.client("secretsmanager", region_name=self.__get_current_region())
+        try:
+            secret = client.get_secret_value(SecretId=secret_identifier)
+            return json.loads(secret["SecretString"])
+        except ClientError as e:
+            raise Exception("Unable to access secret store")
```
atarassov-ttd, line 50: This can swallow important information about why access failed.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+        
+    def __add_defaults(self, configs):
+        configs.setdefault("enclave_memory_mb", 24576)
```
atarassov-ttd, line 53: Where are these default coming from? Why are they reasonable?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+
+    def _get_secret(self, secret_identifier):
+        client = boto3.client("secretsmanager", region_name=self.__get_current_region())
+        try:
+            secret = client.get_secret_value(SecretId=secret_identifier)
+            return json.loads(secret["SecretString"])
+        except ClientError as e:
+            raise Exception("Unable to access secret store")
+        
+    def __add_defaults(self, configs):
```
atarassov-ttd, line 52: Should this be a static method?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+        configs.setdefault("enclave_memory_mb", 24576)
+        configs.setdefault("enclave_cpu_count", 6)
+        configs.setdefault("debug_mode", False)
+        return configs
+
+    def __setup_vsockproxy(self, log_level):
+        thread_count = int((multiprocessing.cpu_count() + 1) // 2)
+        log_level = log_level
```
atarassov-ttd, line 60: Redundant line?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+        try:
+            subprocess.Popen(["/usr/bin/vsockpx", "-c", "/etc/uid2operator/proxy.yaml", "--workers", str(thread_count), "--log-level", log_level, "--daemon"])
```
atarassov-ttd, line 62: This launches process in background without waiting for command to finish. If proxy fails to daemonise for some reason, you may not know.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+        except Exception as e:
+            print(f"Failed to start config server: {e}")
+
+    def __run_socks_proxy(self, log_level):
+        subprocess.Popen(["sockd", "-d"]) 
+
+    def __get_secret_name_from_userdata(self):
+        token = self.__get_aws_token()
+        user_data_url = "http://169.254.169.254/latest/user-data"
+        user_data_response = requests.get(user_data_url, headers={"X-aws-ec2-metadata-token": token})
```
atarassov-ttd, line 87: What if this fails?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+        user_data = user_data_response.text
+        identity_scope = open("/opt/uid2operator/identity_scope.txt").read().strip()
```
atarassov-ttd, line 89: Don't you need to close the file?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+        default_name = "{}-operator-config-key".format(identity_scope.lower())
+        hardcoded_value = "{}_CONFIG_SECRET_KEY".format(identity_scope.upper())
+        match = re.search(rf'^export {hardcoded_value}="(.+?)"$', user_data, re.MULTILINE)
+        return match.group(1) if match else default_name
+
+    def _setup_auxilaries(self):
+        hostname = os.getenv("HOSTNAME", default=os.uname()[1])
+        file_path = "HOSTNAME"
```
atarassov-ttd, line 97: Why is this necessary?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+            "--cpu-count", self.config['enclave_cpu_count'],
+            "--enclave-cid", 42,
+            "--enclave-name", "uid2operator"
+        ]
+        if self.config['debug']:
+            command+=["--debug-mode", "--attach-console"]
+        subprocess.run(command, check=True)
+
+    def cleanup(self):
+        describe_output = subprocess.check_output(["nitro-cli", "describe-enclaves"], text=True)
```
atarassov-ttd, line 144: Why not terminate enclave by name? Or terminate all enclaves?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+        enclaves = json.loads(describe_output)
+        enclave_id = enclaves[0].get("EnclaveID") if enclaves else None
```
atarassov-ttd, line 146: This just kills the first enclave from the list without consideration to whether it is the right one.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,179 @@
...
+            print(f"Failed to shut down {process_name}: {e}")
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-o", "--operation", required=False)
+    args = parser.parse_args()
+    ec2 = EC2()
+    if args.operation and args.operation == "stop":
+        ec2.cleanup()
+        [ec2.kill_process(process) for process in ["vsockpx", "sockd", "vsock-proxy", "nohup"]]
```
atarassov-ttd, line 177: Why is `nohup` in this list? This could potentially kill completely unrelated processes. And why are there both `vsockpx` and `vsock-proxy` in the list?

scripts/confidential_compute.py
```diff
@@ -0,0 +1,73 @@
...
+            SecretNotFoundException: Points to public documentation
+        """
+        pass
+
+    def validate_operator_key(self, secrets):
+        """
+        Validates operator key if following new pattern. Ignores otherwise
+        """
+        api_token = secrets.get('api_token', None)
+        pattern = r"^(UID2|EUID)-.\-(I|P)-\d+-\*$"
```
atarassov-ttd, line 24: Are you sure all operator keys follow this pattern?

scripts/confidential_compute.py
```diff
@@ -0,0 +1,73 @@
...
+        if bool(re.match(pattern, api_token)):
+            if secrets.get('debug_mode', False) or secrets.get('environment') == 'integ':
```
atarassov-ttd, line 26: Why are operator keys forced to integ in debug mode? If debug mode is allowed only for integ, there should be an explicit check for that.

scripts/confidential_compute.py
```diff
@@ -0,0 +1,73 @@
...
+                if api_token.split('-')[2] != 'P':
+                    raise Exception("Operator key does not match the environment")
+        return True
+
+    def validate_connectivity(self, config):
+        """
+        Validates core/optout is accessible. 
+        """
+        try:
+            core_ip = socket.gethostbyname(urlparse(config['core_base_url']).netloc)
```
atarassov-ttd, line 39: What is this for?

scripts/confidential_compute.py
abuabraham-ttd: If it doesn't; we ignore since it is an old key

scripts/confidential_compute.py
abuabraham-ttd: Ok. Can update.

scripts/aws/ec2.py
abuabraham-ttd: we only run 1 enclave on a host,  right? 

can add validation if needed

scripts/aws/ec2.py
abuabraham-ttd: ported from https://github.com/IABTechLab/uid2-operator/blob/main/scripts/aws/stop.sh

scripts/confidential_compute.py
abuabraham-ttd: to provide a meaningful error . 

f"Failed to reach required URLs. Consider enabling {core_ip}, {optout_ip} in the egress firewall."

scripts/confidential_compute.py
```diff
@@ -0,0 +1,83 @@
...
+            core_url = config["core_base_url"]
+            optout_url = config["optout_base_url"]
+            core_ip = self.__resolve_hostname(core_url)
+            requests.get(core_url, timeout=5)
+            optout_ip = self.__resolve_hostname(optout_url)
+            requests.get(optout_url, timeout=5)
+
+        except (requests.ConnectionError, requests.Timeout) as e:
+            raise Exception(
+                f"Failed to reach required URLs. Consider enabling {core_ip}, {optout_ip} in the egress firewall."
```
Ian-Nara, line 65: is it worth distinguishing here between DNS resolution failure and the actual request?

scripts/aws/entrypoint.sh
```diff
@@ -80,6 +80,7 @@ else
   exit 1
 fi
 
+# DO WE NEED THIS? do we expect customers to change URL?
```
Ian-Nara, line 83: I've never heard of it being done before.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,214 @@
...
+            "--enclave-name", "uid2operator"
+        ]
+        if self.configs["debug_mode"]:
+            command += ["--debug-mode", "--attach-console"]
+        subprocess.run(command, check=True)
+
+    def cleanup(self) -> None:
+        """Terminates the Nitro Enclave and auxiliary processes."""
+        try:
+            describe_output = subprocess.check_output(["nitro-cli", "describe-enclaves"], text=True)
```
Ian-Nara, line 178: why not just `nitro-cli terminate-enclave --all`

scripts/confidential_compute.py
```diff
@@ -0,0 +1,83 @@
...
+        api_token = secrets.get("api_token")
+        if not api_token:
+            raise ValueError("API token is missing from the configuration.")
+
+        pattern = r"^(UID2|EUID)-.\-(I|P)-\d+-\*$"
+        if re.match(pattern, api_token):
+            env = secrets.get("environment", "").lower()
+            debug_mode = secrets.get("debug_mode", False)
+            expected_env = "I" if debug_mode or env == "integ" else "P"
+            if api_token.split("-")[2] != expected_env:
```
Ian-Nara, line 41: also throw error if `debug_mode` and `env == prod`? Before attestation failure happens.

abuabraham-ttd: Yep, that can be added as another check 👍

scripts/confidential_compute.py
abuabraham-ttd: Egress rules are writen based on IP and not DNS. If a company implements strict egress rule (for security) they usually will need these IP's to be enabled.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,212 @@
...
+            But can be added in future for tracibility on debug
+            """
+            print(f"Error writing hostname: {e}")
+        self.configs = self._get_secret(self.__get_secret_name_from_userdata())
+        log_level = 3 if self.configs["debug_mode"] else 1
+        self.__setup_vsockproxy(log_level)
+        self.__run_config_server(log_level)
+        self.__run_socks_proxy(log_level)
+
+    def _validate_auxiliaries(self) -> None:
```
sunnywu, line 140: _auxiliaries_ is meaningless - just say Config Server?

for every function that does validation, add some comments to say what you are testing as std out description.

PR #1175, opened 2024-11-20
Cbc UI d2 4454 cloud encryption operator

cody-constine-ttd:


src/main/java/com/uid2/operator/reader/RotatingCloudEncryptionKeyApiProvider.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @Override
+    public long loadContent(JsonObject metadata) throws Exception {
+        return apiStoreReader.loadContent(metadata, "cloudEncryptionKeys");
+    }
+
+    @Override
+    public long getVersion(JsonObject metadata) {
+        return Instant.now().getEpochSecond();
```
Ian-Nara, line 43: why isn't it metadata.getLong("version") ?

src/main/resources/com.uid2.core/test/cloud_encryption_keys/cloud_encryption_keys.json
```diff
@@ -0,0 +1,73 @@
+[ {
```
Ian-Nara, line 1: Indentation format like this: https://github.com/IABTechLab/uid2-admin/blob/main/src/main/resources/localstack/s3/core/keys/keys.json?

src/main/java/com/uid2/operator/reader/RotatingCloudEncryptionKeyApiProvider.java
cody-constine-ttd: Because this is using an API to get the data not a file, so there is no metadata to look the version up in. So I just used the time here to force a refresh everytime

src/test/java/com/uid2/operator/ApiStoreReaderTest.java
```diff
@@ -0,0 +1,103 @@
...
+                    .hasMessageContaining("No array found in the contents");
+        }
+
+        @Test
+        void loadContentSuccessfullyLoadsData() throws Exception {
+            JsonObject contents = new JsonObject()
+                    .put(dataType, new JsonArray().add("value1").add("value2"));
+
+            List<TestData> expectedData = Arrays.asList(new TestData("value1"), new TestData("value2"));
+            when(mockParser.deserialize(any(InputStream.class)))
```
thomasm-ttd, line 73: This test doesn't actually test if the input "contents" are used to return the result. It is only testing that loadContent will store the results of the parser.deserialize in the snapshot. I could change the contents array to have 3 values and the test would still pass.
You should validate properties of the any(InputStream.class) parameter.

thomasm-ttd (commented):
Small comments.

src/main/java/com/uid2/operator/reader/ApiStoreReader.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public long loadContent(JsonObject contents, String dataType) throws IOException {
+        if (contents == null) {
+            throw new IllegalArgumentException(String.format("No contents provided for loading data type %s, cannot load content", dataType));
+        }
+
+        try {
+            JsonArray dataArray = contents.getJsonArray(dataType);
+            if (dataArray == null) {
+                throw new IllegalArgumentException("No array found in the contents");
```
thomasm-ttd, line 39: Would it be useful to include the dataType in the error message?

Ian-Nara (approved):

PR #1191, opened 2024-12-09
Removing start, stop with ec2.py, adding validations

abuabraham-ttd:
Changes enclave start script to Python, to improve readability, remove duplications, and add more validations while starting enclave. 

- Removed unused configurations 
   HOSTNAME, IDENTITY_SCOPE, CORE + OPTOUT_URLS fetched from userdata
- Removes duplications in user-data parsing to get secret manger key name
- Removes unused default cpu, memory and added values from nitro-enclave config to get right value
- Remove read_allocation, update_allocation as it is unused
- Adds typing to config_server, to improve readability
- Kills only valid processes, added flask as it was missing earlier 
- Requires core, optout urls to be specified in secret manager

Testing

Code is tested by creating an EC2 instance using "ami-0df2894e3f4971bda" (the recently published AMI) and copying the python files onto the instance and starting with python files. 

```
python3 -m venv /opt/uid2operator/init
nano requirements.txt and copy the requirements.txt there 
/opt/uid2operator/init/bin/pip install -r requirements.txt 
nano /opt/uid2operator/confidential_compute.py copy and save contents 
nano /opt/uid2operator/ec2.py copy and save contents
nano /etc/systemd/system/uid2operator.service update contents and save
sudo systemctl daemon-reload
sudo systemctl restart uid2operator

and

curl http://127.0.0.1:80/ops/healthcheck

```

You can change the values in secret manager to see errors thrown based on validations added 

Sample run

```
Running in us-east-1
Fetched configs from uid2-config-stack-tjm-unvalidate-eif-test1
Validated https://core-integ.uidapi.com/ matches other config parameters
Validated https://optout-integ.uidapi.com/ matches other config parameters
Validated operator key matches environment
Validated connectivity to https://core-integ.uidapi.com/
Validated connectivity to https://optout-integ.uidapi.com/
Completed static validation of confidential compute config values
Running command: /usr/bin/vsockpx -c /etc/uid2operator/proxy.yaml --workers 5 --log-level 3 --daemon
Running command: ./bin/flask run --host 127.0.0.1 --port 27015
Running command: sockd -D
Connecting to config server, attempt 1 failed with ConnectionError: HTTPConnectionPool(host='127.0.0.1', port=27015): Max retries exceeded with url: /getConfig (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7efebaadfca0>: Failed to establish a new connection: [Errno 111] Connection refused'))
Config server is reachable
Connectivity check to config server passes
Running in debug_mode
Running command: nitro-cli run-enclave --eif-path /opt/uid2operator/uid2operator.eif --memory 24576 --cpu-count 6 --enclave-cid 42 --enclave-name uid2operator --debug-mode --attach-console
Start allocating memory...
```


scripts/aws/ec2.py
```diff
@@ -0,0 +1,216 @@
...
+        """
+        Starts the SOCKS proxy service.
+        """
+        command = ["sockd", "-D"]
+        self.run_command(command)
+
+    def __get_secret_name_from_userdata(self) -> str:
+        """Extracts the secret name from EC2 user data."""
+        token = self.__get_aws_token()
+        user_data_url = "http://169.254.169.254/latest/user-data"
```
sunnywu, line 122: maybe an optional future improvement tix: do you want to not hardcode the path/IP address inside various functions but to set it as global/static variables somewhere up the top?

i suppose keeping it as is helps reviewing when comparing the new codes with old bash codes

scripts/aws/ec2.py
```diff
@@ -0,0 +1,220 @@
...
+        except Exception as e:
+            print(f"Error killing process '{process_name}': {e}")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(description="Manage EC2-based confidential compute workflows.")
+    parser.add_argument("-o", "--operation", choices=["stop", "start"], default="start", help="Operation to perform.")
+    args = parser.parse_args()
+    ec2 = EC2()
+    if args.operation == "stop":
```
sunnywu, line 216: the old start.sh would run terminate enclave not just during stop.sh - should this be replicated here too?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,220 @@
...
+        default_name = f"{identity_scope.lower()}-operator-config-key"
+        hardcoded_value = f"{identity_scope.upper()}_CONFIG_SECRET_KEY"
+        match = re.search(rf'^export {hardcoded_value}="(.+?)"$', user_data, re.MULTILINE)
+        return match.group(1) if match else default_name
+
+    def _setup_auxiliaries(self) -> None:
+        """Sets up the necessary auxiliary services and configuration."""
+        self.configs = self._get_secret(self.__get_secret_name_from_userdata())
+        self.validate_configuration()
+        log_level = 3 if self.configs["debug_mode"] else 1
```
sunnywu, line 143: is this right translation from the old bash script:

```VSOCK_LOG_LEVEL=${VSOCK_LOG_LEVEL:-3}```
?
(granted don't quite understand this bash script)

File: scripts/aws/ec2.py [hunk 4]
```diff
@@ -0,0 +1,220 @@
...
+
+        default_name = f"{identity_scope.lower()}-operator-config-key"
+        hardcoded_value = f"{identity_scope.upper()}_CONFIG_SECRET_KEY"
+        match = re.search(rf'^export {hardcoded_value}="(.+?)"$', user_data, re.MULTILINE)
+        return match.group(1) if match else default_name
+
+    def _setup_auxiliaries(self) -> None:
+        """Sets up the necessary auxiliary services and configuration."""
+        self.configs = self._get_secret(self.__get_secret_name_from_userdata())
+        self.validate_configuration()
```
sunnywu, line 142: shouldn't this be a standalone step in run_compute function? otherwise it looks hidden inside  ```_setup_auxiliaries``` ?

scripts/aws/start.sh
```diff
@@ -1,124 +0,0 @@
...
-}
-
-terminate_old_enclave
-config_aws
-read_allocation
-# update_allocation
-setup_vsockproxy
-setup_dante
-run_config_server
-run_enclave
```
sunnywu, line 122: i thought this is very clear in terms of what are the actual steps and now most of this is grouped under ```_setup_auxiliaries``` in py script which isn't necessarily as just easy to read what are the steps comparing to this. 

i don't have any suggestion how to improve but food for thought.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,220 @@
...
+            return response.json()["region"]
+        except requests.RequestException as e:
+            raise RuntimeError(f"Failed to fetch region: {e}")
+        
+    def __validate_aws_specific_config(self, secret):
+        if "enclave_memory_mb" in secret or "enclave_cpu_count" in secret:
+            max_capacity = self.__get_max_capacity()
+            for key in ["enclave_memory_mb", "enclave_cpu_count"]:
+                if int(secret.get(key, 0)) > max_capacity.get(key):
+                    raise ValueError(f"{key} value ({secret.get(key, 0)}) exceeds the maximum allowed ({max_capacity.get(key)}).")
```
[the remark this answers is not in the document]
abuabraham-ttd: yeah, makes sense.

scripts/aws/ec2.py
abuabraham-ttd: if VSOCK_LOG_LEVEL ? then VSOCK_LOG_LEVEL else 3

there is no way to set VSOCK_LOG_LEVEL though. 

So I added if debug_mode ? then set to 3 otherwise 1. 

In future ticket, we will be adding more tracing on debug mode. 

Plan is to evaluate adding 

echo 1 > /sys/kernel/debug/tracing/events/vsock/enable
echo 1 > /sys/kernel/debug/tracing/tracing_on

scripts/aws/ec2.py
abuabraham-ttd: Dont think it is needed. 

We manage enclaves through systemd. systemd restart will do stop/start (as seen in uid2operator.service)

scripts/aws/ec2.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _setup_auxiliaries(self) -> None:
+        """Sets up the necessary auxiliary services and configuration."""
+        self.configs = self._get_secret(self.__get_secret_name_from_userdata())
+        self.validate_configuration()
+        log_level = 3 if self.configs["debug_mode"] else 1
+        self.__setup_vsockproxy(log_level)
+        self.__run_config_server()
+        self.__run_socks_proxy()
+        time.sleep(5)  #TODO: Change to while loop if required.
```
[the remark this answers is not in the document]
abuabraham-ttd: or is this to wait for config server to startup so you could do validations? i 

Yes. we can "Change to while loop if required. " as mentioned in comment.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,235 @@
...
+            response = requests.get(metadata_url, headers=headers, timeout=2)
+            response.raise_for_status()
+            return response.json()["region"]
+        except requests.RequestException as e:
+            raise RuntimeError(f"Failed to fetch region: {e}")
+        
+    def __validate_aws_specific_config(self, secret):
+        if "enclave_memory_mb" in secret or "enclave_cpu_count" in secret:
+            max_capacity = self.__get_max_capacity()
+            min_capacity = {"enclave_memory_mb": 10000, "enclave_cpu_count" : 4 }
```
Ian-Nara, line 57: should be a multiple of 1024. Was 10Gb successful in prod to hold all the optouts & salts? I got OOM with 9Gb.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,235 @@
...
+        
+        region = self.__get_current_region()
+        print(f"Running in {region}")
+        try:
+            client = boto3.client("secretsmanager", region_name=region)
+        except Exception as e:
+            raise RuntimeError("Please use IAM instance profile for your instance that has permission to access Secret Manager")
+        try:
+            secret = json.loads(client.get_secret_value(SecretId=secret_identifier)["SecretString"])
+            self.__validate_aws_specific_config(secret)
```
Ian-Nara, line 83: should we add defaults before validating the secret? what if the max capacity is below our minimum?

scripts/aws/ec2.py
abuabraham-ttd: No, just a random value I put. Was looking for inputs. What is the min we have tested successfully?

scripts/aws/ec2.py
abuabraham-ttd: max capacity is the max allocated to nitro-encalve service, if it is below our minimum, it fails as expected

scripts/aws/start.sh
abuabraham-ttd: abstracted to `_setup_auxiliaries` which has

self.__setup_vsockproxy(log_level)
self.__run_config_server()
self.__run_socks_proxy()

and run_compute to run compute. 

I went on that route, as developers can easily compare and understand differences between Cloud provider implementation. 

As in, _setup_auxiliaries will setup all required processes for starting enclave. 

If you need to know what Azure needs, just look into _setup_auxiliaries in Azure.py 

This also enables the consistently phased rollout

scripts/aws/ec2.py
Ian-Nara: I don't think we have gone below 20Gb, but I am testing lower, down to 16Gb instance

Ian-Nara (approved):

scripts/aws/ec2.py
```diff
@@ -0,0 +1,235 @@
...
+        """Main execution flow for confidential compute."""
+        secret_manager_key = self.__get_secret_name_from_userdata()
+        self.configs = self._get_secret(secret_manager_key)
+        print(f"Fetched configs from {secret_manager_key}")
+        self.validate_configuration()
+        self._setup_auxiliaries()
+        self._validate_auxiliaries()
+        command = [
+            "nitro-cli", "run-enclave",
+            "--eif-path", "/opt/uid2operator/uid2operator.eif",
```
cody-constine-ttd, line 185: would be helpful for debugging to have the ability to pass this in.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,235 @@
...
+        ]
+        if self.configs["debug_mode"]:
+            print("Running in debug_mode")
+            command += ["--debug-mode", "--attach-console"]
+        self.run_command(command)
+
+    def cleanup(self) -> None:
+        """Terminates the Nitro Enclave and auxiliary processes."""
+        try:
+            describe_output = subprocess.check_output(["nitro-cli", "describe-enclaves"], text=True)
```
cody-constine-ttd, line 199: There is actually no need to do this. You can terminate by enclave name

scripts/aws/ec2.py
```diff
@@ -0,0 +1,235 @@
...
+        """Sets up the vsock tunnel, socks proxy and flask server"""
+        log_level = 3 if self.configs["debug_mode"] else 1
+        self.__setup_vsockproxy(log_level)
+        self.__run_config_server()
+        self.__run_socks_proxy()
+
+    def _validate_auxiliaries(self) -> None:
+        """Validates connection to flask server direct and through socks proxy."""
+        proxy = "socks5://127.0.0.1:3306"
+        config_url = "http://127.0.0.1:27015/getConfig"
```
cody-constine-ttd, line 152: We should probably make a config for stuff like this. Much easier to change/see what these values are in one place. Especially for config url and port because you set in two places.

scripts/aws/ec2.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def run_compute(self) -> None:
+        """Main execution flow for confidential compute."""
+        secret_manager_key = self.__get_secret_name_from_userdata()
+        self.configs = self._get_secret(secret_manager_key)
+        print(f"Fetched configs from {secret_manager_key}")
+        self.validate_configuration()
+        self._setup_auxiliaries()
+        self._validate_auxiliaries()
+        command = [
```
cody-constine-ttd, line 183: Feels like nitro start should be its own function

scripts/aws/ec2.py
abuabraham-ttd: why? 

We need to build an EIF and register the EIF in core. Why not just use the ami build along with EIF? otherwise we have to copy the registered EIF and run that by SSH ing into the host right?

scripts/aws/ec2.py
sunnywu: just my 2 cents - it doesn't need to be as fine grained  "socks5://{AuxilaryConfig.LOCALHOST}:{AuxilaryConfig.SOCKS_PORT}" but just define ```SOCK5_PROXY_URL = ""socks5://127.0.0.1:3306"``` under AuxilaryConfig. at least if everything under AuxilaryConfig we can catch inconsistencies more easily already.

scripts/aws/ec2.py
sunnywu: _there is no way to set VSOCK_LOG_LEVEL though._

Are you saying in the old start.h, we never really set VSOCK_LOG_LEVEL but rely on the default 3 value [here](https://github.com/IABTechLab/uid2-operator/pull/1191/files#diff-b1f9a752648fa0d8bc1d5b4041260c0e787942dc73cb271c8344e40c4f2b05fdL87)?  

Anyway i'm saying your latest vsock debug level default  is setting to 1 now unless someone explicitly turn on debug_mode config. Which may look a bit different to how  VSOCK_LOG_LEVEL is set previously but i dunno if VSOCK_LOG_LEVEL was ever set to a different value other than the default 3. 

So i think your latest change is ok, but probably would ask @cody-constine-ttd or @thomasm-ttd for confirmation

scripts/aws/ec2.py
```diff
@@ -0,0 +1,251 @@
...
+            default_capacity = self.__get_max_capacity()
+            configs.setdefault("enclave_memory_mb", default_capacity["enclave_memory_mb"])
+            configs.setdefault("enclave_cpu_count", default_capacity["enclave_cpu_count"])
+            configs.setdefault("debug_mode", False)
+            return configs
+        
+        region = self.__get_current_region()
+        print(f"Running in {region}")
+        try:
+            client = boto3.client("secretsmanager", region_name=region)
```
sunnywu, line 101: if this is successful - the logs should print "Secret Manager accessed successfully"?

scripts/aws/ec2.py
```diff
@@ -0,0 +1,251 @@
...
+            max_capacity = self.__get_max_capacity()
+            min_capacity = {"enclave_memory_mb": 11000, "enclave_cpu_count" : 2 }
+            for key in ["enclave_memory_mb", "enclave_cpu_count"]:
+                if int(secret.get(key, 0)) > max_capacity.get(key):
+                    raise ValueError(f"{key} value ({secret.get(key, 0)}) exceeds the maximum allowed ({max_capacity.get(key)}).")
+                if min_capacity.get(key) > int(secret.get(key, 10**9)):
+                    raise ValueError(f"{key} value ({secret.get(key, 0)}) needs to be higher than the minimum required ({min_capacity.get(key)}).")
+                
+    def _get_secret(self, secret_identifier: str) -> AWSConfidentialComputeConfig:
+        """Fetches a secret value from AWS Secrets Manager and adds defaults"""
```
sunnywu, line 88: just a suggestion - I think for most comments like this in this ec2.py, you probably should could log it out as well so whoever runs it knows which stage of validation/startup it is on. If i run this script, i would prefer to see all steps done in this script

abuabraham-ttd: we do these where the checks are implemented (in confidential_compute.py)

scripts/aws/ec2.py
abuabraham-ttd: we do these where the checks are implemented (in confidential_compute.py)

scripts/aws/ec2.py
abuabraham-ttd: see the sample run output I have added in desc

scripts/aws/ec2.py
```diff
@@ -0,0 +1,259 @@
...
+    args = parser.parse_args()
+    try:
+        ec2 = EC2()
+        if args.operation == "stop":
+            ec2.cleanup()
+        else:
+            ec2.run_compute()
+    except ConfidentialComputeStartupException as e:
+        print("Failed starting up Confidential Compute. Please checks the logs for errors and retry \n", e)
+    except Exception as e:
```
abuabraham-ttd, line 257: The goal is; catch all failures we have seen in ConfidentialComputeStartupException. This points to public doc, something like;


Screenshot 2024-12-19 at 4 59 33 PM

And we always add validations for any seen error so it is upto date. If an unseen error, talk to UID2 support

scripts/aws/EUID_CloudFormation.template.yml
```diff
@@ -86,6 +100,12 @@ Metadata:
         default: OPERATOR_KEY provided by EUID Administrator.
       DeployToEnvironment:
         default: EUID environment to deploy to. Prod - production; Integ - integration test.
+      CoreBaseURL:
```
thomasm-ttd, line 103: I would really not like to have these as public parameters on the template - this impacts our public documentation and there are more things participants have to fill in when they launch the template.

abuabraham-ttd: https://thetradedesk.slack.com/archives/C084UCJ957W/p1734475328491369?thread_ts=1734041429.068699&cid=C084UCJ957W 

We decided that `core_base_url` is what needs to be provided (since optout_url returned by core is not used everwhere until that is done, add `optout_url` as well) 

But I think your concern can be addressed for this template, 

we use seperate files for euid, uid. So we know the base url for is only dependent on env. I can check if we can someway to string manipulation in CFN template.

However, this will only work for ec2 ami

pom.xml
```diff
@@ -6,7 +6,7 @@
 
     <groupId>com.uid2</groupId>
     <artifactId>uid2-operator</artifactId>
-    <version>5.43.0</version>
+    <version>5.43.8-alpha-105-SNAPSHOT</version>
```
sunnywu, line 9: should this be reversed or changed to just` 5.43.8`?

scripts/aws/UID_CloudFormation.template.yml
```diff
@@ -65,6 +69,7 @@ Metadata:
         Parameters:
           - APIToken
           - DeployToEnvironment
+          - SkipValidations
```
thomasm-ttd, line 72: This shouldn't be in the UI

cody-constine-ttd (approved):

scripts/confidential_compute.py
```diff
@@ -0,0 +1,141 @@
...
+        """ Validates the paramters specified through configs/secret manager ."""
+        print("Validating configurations provided")
+        def validate_operator_key():
+            """ Validates the operator key format and its environment alignment."""
+            operator_key = self.configs.get("api_token")
+            if not operator_key:
+                raise ValueError("API token is missing from the configuration.")
+            pattern = r"^(UID2|EUID)-.\-(I|P|L)-\d+-.*$"
+            if re.match(pattern, operator_key):
+                env = self.configs.get("environment", "").lower()
```
sunnywu, line 32: should we explicitly check the value of `environment` config - it seems it always assume it's either integ/prod but could it be overwritten with invalid value?

abuabraham-ttd: added

pom.xml
abuabraham-ttd: I did not update it, maybe release automatically changes it? @cody-constine-ttd @thomasm-ttd

PR #1260, opened 2025-01-09
UID2-4719 change azure cc starting process

clarkxuyang:


scripts/aws/ec2.py
```diff
@@ -51,7 +51,8 @@ def get_meta_url(cls) -> str:
 class EC2(ConfidentialCompute):
 
     def __init__(self):
-        super().__init__()
+        self.configs: AWSConfidentialComputeConfig = {}
```
abuabraham-ttd, line 54: Do we need this here? self.configs {} is defined in base class

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -0,0 +1,176 @@
...
+        self.__wait_for_sidecar()
+        self.__run_operator()
+
+    def _setup_auxiliaries(self) -> None:
+        """ Sets up auxiliary processes required for confidential computing. """
+        pass
+
+    def _validate_auxiliaries(self) -> None:
+        """ Validates auxiliary services are running."""
+        pass
```
abuabraham-ttd, line 164: shouldn't __wait_for_sidecar be here?

scripts/confidential_compute.py
```diff
@@ -5,6 +5,7 @@
 from abc import ABC, abstractmethod
 from typing import TypedDict, NotRequired, get_type_hints
 import subprocess
+import logging
```
abuabraham-ttd, line 8: 👍

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -0,0 +1,176 @@
...
+import logging
+
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+from confidential_compute import ConfidentialCompute, ConfidentialComputeConfig, MissingConfig, ConfidentialComputeStartupException 
+from azure.identity import DefaultAzureCredential, CredentialUnavailableError
+from azure.keyvault.secrets import SecretClient
+from azure.core.exceptions import ResourceNotFoundError, HttpResponseError
+
+class AzureEntryPoint(ConfidentialCompute):
+
```
abuabraham-ttd, line 19: Chage this to Azure? This class name is used to create Errors/point to docs in ConfidentialComputeStartupException

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -0,0 +1,176 @@
...
+    def _set_secret(self, secret_identifier: str = None):
+        try:
+            credential = DefaultAzureCredential()
+            kv_URL = f"https://{AzureEntryPoint.kv_name}.vault.azure.net"
+            secret_client = SecretClient(vault_url=kv_URL, credential=credential)
+            secret = secret_client.get_secret(AzureEntryPoint.secret_name)
+            # print(f"Secret Value: {secret.value}")
+            self.configs["api_token"] = secret.value
+
```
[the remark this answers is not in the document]
abuabraham-ttd: IMO; 

_set_secret should be setting ConfidentialComputeConfig (which contains all the possible customer inputs to create Confidential Compute env for running operators) . That's why I used _get_secret() -> ConfidentialComputeConfig 

If we use _set_secret -> None, then there is no way to ensure/understand what values it sets.

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -0,0 +1,176 @@
...
+        except CredentialUnavailableError as auth_error:
+            logging.error(f"Read operator key, authentication error: {auth_error}")
+            raise
```
abuabraham-ttd, line 54: raise custom exception here? That points to public doc? This is the same as Missing Instance Profile in AWS

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -0,0 +1,176 @@
...
+
+        except ResourceNotFoundError as not_found_error:
+            logging.error(f"Read operator key, secret not found: {AzureEntryPoint.secret_name}. Error: {not_found_error}")
+            raise
```
abuabraham-ttd, line 58: raise custom exception here? That points to public doc?  ConfigNotFound

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -0,0 +1,176 @@
...
+
+        except HttpResponseError as http_error:
+            logging.error(f"Read operator key, HTTP error occurred: {http_error}")
+            raise
```
abuabraham-ttd, line 62: When would we have this? Is it tied to vnet creation? If Container Group does not have n/w configured? If so, another custom exception for Azure that points to doc, and explaining what is wrong

scripts/aws/ec2.py
clarkxuyang: in base class, self.configs is defined as ConfidentialComputeConfig, to use extended field in AWSConfidentialComputeConfig in EC2 class, we need to define self.config as AWSConfidentialComputeConfig

scripts/azure-cc/azureEntryPoint.py
clarkxuyang: The class name of "Azure", or "EC2" looks not good, let's discuss

scripts/azure-cc/azureEntryPoint.py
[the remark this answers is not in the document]
clarkxuyang: We can change set_secret to set_ ConfidentialComputeConfig, and use it as orch point to call other private functions to set individual field of ConfidentialComputeConfig

scripts/azure-cc/azureEntryPoint.py
sunnywu: Any decision made here?

clarkxuyang: updated xxxEntryPoint

scripts/gcp-oidc/gcp.py
```diff
@@ -37,12 +35,9 @@ def _get_secret(self, secret_identifier=None) -> ConfidentialComputeConfig:
             secret_version_name = f"{os.getenv("API_TOKEN_SECRET_NAME")}"
             response = client.access_secret_version(name=secret_version_name)
             secret_value = response.payload.data.decode("UTF-8")
-        except (PermissionDenied, DefaultCredentialsError) as e:
-            raise MissingInstanceProfile(self.__class__.__name__, str(e))
-        except NotFound:
-            raise ConfigNotFound(self.__class__.__name__, f"Secret Manager {os.getenv("API_TOKEN_SECRET_NAME")}")
```
clarkxuyang, line 43: API_TOKEN_SECRET_NAME is already checked in line 33

scripts/gcp-oidc/gcp.py
```diff
@@ -37,12 +35,9 @@ def _get_secret(self, secret_identifier=None) -> ConfidentialComputeConfig:
             secret_version_name = f"{os.getenv("API_TOKEN_SECRET_NAME")}"
             response = client.access_secret_version(name=secret_version_name)
             secret_value = response.payload.data.decode("UTF-8")
-        except (PermissionDenied, DefaultCredentialsError) as e:
-            raise MissingInstanceProfile(self.__class__.__name__, str(e))
```
abuabraham-ttd, line 41: MissingInstanceProfile 

is when you dont have a Service Account attached, we need that

scripts/gcp-oidc/gcp.py
abuabraham-ttd: API_TOKEN_SECRET_NAME is the name passed in through env variables. That may/maynot be correct, this is checking if we are unable to access the secret with name API_TOKEN_SECRET_NAME

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -0,0 +1,172 @@
...
+            "-Dlogback.configurationFile=/app/conf/logback.xml",
+            f"-Dvertx-config-path={AzureEntryPoint.FINAL_CONFIG}",
+            "-jar", 
+            f"{AzureEntryPoint.jar_name}-{AzureEntryPoint.jar_version}.jar"
+        ]
+        logging.info("-- starting java operator application")
+        self.run_command(java_command, separate_process=False)
+
+    def _setup_auxiliaries(self):
+        logging.info("Waiting for sidecar ...")
```
abuabraham-ttd, line 119: should this be in validate _auxiliaries ?we are setting it up elsewhere

scripts/confidential_compute.py
```diff
@@ -77,15 +86,15 @@ def validate_url(url_key, environment):
             parsed_url = urlparse(self.configs[url_key])
             if parsed_url.scheme != 'https' and parsed_url.path:
                 raise InvalidConfigValue(self.__class__.__name__, url_key)
-            print(f"Validated {self.configs[url_key]} matches other config parameters")
+            logging.info(f"Validated {self.configs[url_key]} matches other config parameters")
```
abuabraham-ttd, line 90: Thank you for changing the prints to logging everywhere

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -0,0 +1,176 @@
...
+            self.configs["core_base_url"] = jdata["core_attest_url"]
+            self.configs["optout_base_url"] = jdata["optout_api_uri"]
+
+    def __run_operator(self):
+
+        # Start the operator
+        os.environ["azure_vault_name"] = AzureEntryPoint.kv_name
+        os.environ["azure_secret_name"] = AzureEntryPoint.secret_name
+
+        java_command = [
```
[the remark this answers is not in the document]
clarkxuyang: Good idea. I will leave it to the config optimization task, and focus this PR for improve enclave test and debugging.

abuabraham-ttd (approved):

PR #1289, opened 2025-01-16
Remove and replace additional operations in enclave start

abuabraham-ttd:


scripts/aws/entrypoint.sh
```diff
@@ -18,82 +18,66 @@ ifconfig lo 127.0.0.1
...
+  FILTER=${FILTER%, }'}'
+  jq "${FILTER}" "${PARAMETERIZED_CONFIG}" > "${PARAMETERIZED_CONFIG}.tmp" && mv "${PARAMETERIZED_CONFIG}.tmp" "${PARAMETERIZED_CONFIG}"
+}
+
+build_operator_config() {
+  CORE_BASE_URL=$(jq -r ".core_base_url" < "${PARAMETERIZED_CONFIG}")
+  OPTOUT_BASE_URL=$(jq -r ".optout_base_url" < "${PARAMETERIZED_CONFIG}")
+  DEPLOYMENT_ENVIRONMENT=$(jq -r ".environment" < "${OPERATOR_CONFIG}")
+  DEBUG_MODE=$(jq -r ".debug_mode" < "${OPERATOR_CONFIG}")
+
```
abuabraham-ttd, line 50: we need to ensure CORE_BASE_URL is not set to random core, as this will result in operator sending token + pcr0 to overridden malicious core. 

Basically, updates the logic to always replace URL in config. If prod, use default prod URL otherwise allow override. 

This check is not done in GCP, as we attest GCP confidential compute by calling GCP endpoints. Please correct if wrong @atarassov-ttd

scripts/aws/entrypoint.sh
```diff
@@ -18,82 +18,65 @@ ifconfig lo 127.0.0.1
...
+build_parameterized_config() {
+  curl -s -f -o "${PARAMETERIZED_CONFIG}" -x socks5h://127.0.0.1:3305 http://127.0.0.1:27015/getConfig
+  REQUIRED_KEYS=("optout_base_url" "core_base_url" "api_token" "environment")
+  for key in "${REQUIRED_KEYS[@]}"; do
+    if ! jq -e "has(\"${key}\")" "${PARAMETERIZED_CONFIG}" > /dev/null; then
+      echo "Error: Key '${key}' is missing. Please add it to flask config server"
+      exit 1
+    fi
+  done
+  FILTER=$(printf '. | {')
```
abuabraham-ttd, line 35: We do this logic differently in each implementation: In GCP in Java code, in Azure in MAA attestation, and in AWS it was in make_config.py.

scripts/aws/config-server/app.py
```diff
@@ -10,25 +10,6 @@ def get_config():
         with open('/etc/secret/secret-value/config', 'r') as secret_file:
             secret_value = secret_file.read().strip()
             secret_value_json = json.loads(secret_value)
-            secret_value_json["environment"] = secret_value_json["environment"].lower()
```
abuabraham-ttd, line 13: Is this needed now? 

We populate configs after processing in https://github.com/IABTechLab/uid2-operator/blob/9b94fd56999819561a492bcb59f0fdb5143dff80/scripts/aws/ec2.py#L135

scripts/aws/config-server/app.py
```diff
@@ -10,25 +10,6 @@ def get_config():
...
-            if "core_base_url" in secret_value_json:
-                secret_value_json["core_base_url"] = secret_value_json["core_base_url"].lower()
-            if "optout_base_url" in secret_value_json:
-                secret_value_json["optout_base_url"] = secret_value_json["optout_base_url"].lower()
-            if "operator_type" in secret_value_json and secret_value_json["operator_type"].lower() == "public":
-                mount_path = '/etc/config/config-values'
```
abuabraham-ttd, line 19: Is this to populate extra configs outside of Secret manager for eks operator?

scripts/aws/entrypoint.sh
```diff
@@ -7,93 +7,87 @@ LOG_FILE="/home/start.txt"
...
-elif [ "${IDENTITY_SCOPE}" = "EUID" ]; then
-  python3 /app/make_config.py /app/conf/prod-euid-config.json /app/conf/integ-euid-config.json ${OVERRIDES_CONFIG} "$(nproc)" > ${FINAL_CONFIG}
-else
-  echo "Unrecognized IDENTITY_SCOPE ${IDENTITY_SCOPE}"
-  exit 1
-fi
+  #TODO: Remove below logic after remote config management is implemented
+
+  if [[ "$DEPLOYMENT_ENVIRONMENT" != "prod" ]]; then
+    #Allow override of base URL in non-prod environments
```
abuabraham-ttd, line 69: I tried to remove all configs we pass and keep it simple, but there are lot of intricacies 

This bit, can be removed once we complete remote config management, and use optour url returned by core.

scripts/aws/config-server/app.py
Ian-Nara: I think the lower() calls to these is still useful, unless I missed that in the ec2.py

abuabraham-ttd: we ensure it is lower as part of validations

scripts/aws/config-server/app.py
abuabraham-ttd: let us do it outside, while starting flask, read from configmaps

Ian-Nara (approved):

PR #1317, opened 2025-01-24
Configservice Healthmanager Integration

BehnamMozafari:
- Integrate ConfigService with HealthManager
- Instance is set to unhealthy if config is not retrieved from the core endpoint after 3 consecutive attempts
- Instance set back to healthy once config retrieval is successful


src/main/java/com/uid2/operator/service/ConfigService.java
```diff
@@ -1,24 +1,45 @@
...
-        this.configRetriever.setConfigurationProcessor(this::configValidationHandler);
+        this.configRetriever.setConfigurationProcessor(json -> {
+            JsonObject validJson = configValidationHandler(json);
+            failedRetrievalCount.set(0);
+            this.healthComponent.setHealthStatus(true, "config retrieved successfully");
+            return validJson;
+        });
+        this.configRetriever.configStream()
+                .exceptionHandler(e -> {
+                    logger.warn("Exception occurred while retrieving configuration: ", e);
```
jon8787, line 36: Is there any more info we can add here, such as the URL or file where it's trying to retrieve from?

Also a more general question - what does Operator do if it is never able to retrieve the config? Does it still accept requests?

BehnamMozafari: > Is there any more info we can add here, such as the URL or file where it's trying to retrieve from?

with the current implementation, we can't access any metadata from ConfigRetriever, but we can possibly extend ConfigRetriever or use a wrapper to get access to that info - what are your thoughts on doing this?

> what does Operator do if it is never able to retrieve the config? Does it still accept requests?

On startup, if operator can't retrieve the runtime config or retrieves invalid config, it will throw an exception and fail to start. After startup, it will fall back to the last valid config if it retrieves an invalid one.

jon8787: Can you paste example logs where this fails due to a URL it can't reach? Does the URL appear near this exception log? If yes, we can leave it. Otherwise yes - we should consider wrapping ConfigRetriever or similar.

src/main/java/com/uid2/operator/service/ConfigService.java
```diff
@@ -1,24 +1,45 @@
...
     private final ConfigRetriever configRetriever;
     private static final Logger logger = LoggerFactory.getLogger(ConfigService.class);
+    private final HealthComponent healthComponent = HealthManager.instance.registerComponent("config-service");
+    private static final int MAX_FAILURE_COUNT = 3;
+    private final AtomicInteger failedRetrievalCount = new AtomicInteger(0);
 
     private ConfigService(ConfigRetriever configRetriever) {
         this.configRetriever = configRetriever;
-        this.configRetriever.setConfigurationProcessor(this::configValidationHandler);
+        this.configRetriever.setConfigurationProcessor(json -> {
```
sunnywu, line 28: could you add a unit test for this change? mocking not able to receive the config?

src/main/java/com/uid2/operator/service/ConfigService.java
BehnamMozafari: "The timeout period of 800ms has been exceeded while executing GET /operator/config for server null" That's an example of what it would log, will address this in a future pr though, as currently the only config source it will be retrieving from is the core endpoint. We also have similar work planned to update the HTTP Store to validate response status codes, so this would go along with that.

PR #1340, opened 2025-02-04
UID2-4808 Add azure-cc-aks protocol

cYKatherine:


src/main/java/com/uid2/operator/Main.java
```diff
@@ -508,7 +509,10 @@ private AttestationResponseHandler getAttestationTokenRetriever(Vertx vertx, Str
                 break;
             case "azure-cc":
                 LOGGER.info("creating uid core client with azure cc attestation protocol");
-                String maaServerBaseUrl = this.config.getString(Const.Config.MaaServerBaseUrlProp, "https://sharedeus.eus.attest.azure.net");
+                attestationProvider = AttestationFactory.getAzureCCAttestation(maaServerBaseUrl);
+                break;
+            case "azure-cc-aks":
```
cYKatherine, line 514: QQ: We specify `enclavePlatform` in config, however, `azure-cc` and `azure-cc-aks` gets the same attestation provider. Do we still want to separate `azure-cc` and `azure-cc-aks` for `enclavePlatform`? It's quite confusing comapred to `protocol` in operator key, should we potentially make them a same thing in the future?

abuabraham-ttd, line 514: cant we do 

case "azure-cc", "azure-cc-aks":

abuabraham-ttd: I think it is fine since platform is different, attestation is same.

PR #1347, opened 2025-02-04
Update to fix new config errors, change exception name

abuabraham-ttd:


scripts/gcp-oidc/gcp.py
```diff
@@ -32,25 +32,25 @@ def _set_confidential_config(self, secret_identifier=None) -> None:
...
         self.configs["operator_key"] = secret_value
     
     def __populate_operator_config(self, destination):
         target_config = f"/app/conf/{self.configs["environment"].lower()}-config.json"
         shutil.copy(target_config, destination)
         with open(destination, 'r') as file:
             config = file.read()
-        config = config.replace("https://core.uidapi.com", self.configs.get("core_base_url"))
-        config = config.replace("https://optout.uidapi.com", self.configs.get("optout_base_url"))
+        config = config.replace("core.uidapi.com", urlparse(self.configs.get("core_base_url")).netloc)
```
clarkxuyang, line 52: is this suppose to be core-integ.uidapi.com

abuabraham-ttd: no, in GCP config is/(and was before our update) core.uidapi.com . We replace values for integ and prod both. 

This is fine because we attest GCP solution by calling GCP API. Cant spoof in any way. 

Once we have remote-config rollout complete, we can remove the 2 files and just use 1 as well

scripts/azure-cc/deployment/operator.json
```diff
@@ -122,6 +122,10 @@
                 {
                   "name": "DEPLOYMENT_ENVIRONMENT",
                   "value": "[parameters('deploymentEnvironment')]"
+                },
+                {
+                  "name": "SKIP_VALIDATIONS",
+                  "value": "[parameters('skipValidations')]"
```
clarkxuyang, line 128: Do we need to declare skipValidations in the parameter list?

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -143,7 +144,7 @@ def _validate_auxiliaries(self):
                 logging.error(
                     f"Sidecar failed to start after {MAX_RETRIES} attempts. Exiting."
```
clarkxuyang, line 145: Could you help to change to "Unable to detect sidecar running after {MAX_RETRIES} attempts. Exiting." Thanks!

clarkxuyang (approved):

PR #1354, opened 2025-02-05
Fix feat flag listener

BehnamMozafari:
- Fix feature flag listener to read flag from correct location in `feat-flag.json`
- Default feature flag to off if not successfully retrieved from file (e.g. invalid json)
- Made feature flag config retriever file store optional


src/main/java/com/uid2/operator/Main.java
```diff
@@ -288,15 +290,18 @@ private Future<IConfigService> initialiseConfigService() throws Exception {
...
+        ConfigRetriever featureFlagConfigRetriever = ConfigRetriever.create(
                 vertx,
-                new JsonObject()
-                        .put("type", "file")
-                        .put("config", new JsonObject()
-                                .put("path", "conf/feat-flag/feat-flag.json")
-                                .put("format", "json"))
-                        .put(ConfigScanPeriodMsProp, 10000),
-                ""
+                new ConfigRetrieverOptions(
```
jon8787, line 295: is it worth adding a test that would have failed?

src/main/java/com/uid2/operator/Main.java
```diff
@@ -377,12 +382,12 @@ private void run() throws Exception {
     private void setupFeatureFlagListener(ConfigServiceManager manager, ConfigRetriever retriever) {
         retriever.listen(change -> {
             JsonObject newConfig = change.getNewConfiguration();
-            boolean useDynamicConfig = newConfig.getBoolean(Const.Config.RemoteConfigFeatureFlagProp, true);
+            boolean useDynamicConfig = newConfig.getJsonObject("remote_config", new JsonObject()).getBoolean("enabled", false);
```
jon8787, line 385: are we removing the const as well?

src/main/java/com/uid2/operator/Main.java
```diff
@@ -288,15 +290,18 @@ private Future<IConfigService> initialiseConfigService() throws Exception {
...
-                new JsonObject()
-                        .put("type", "file")
-                        .put("config", new JsonObject()
-                                .put("path", "conf/feat-flag/feat-flag.json")
-                                .put("format", "json"))
-                        .put(ConfigScanPeriodMsProp, 10000),
-                ""
+                new ConfigRetrieverOptions(
+                        new JsonObject()
+                                .put("stores", new JsonArray()
```
jon8787, line 297: where is this json structure defined? can we add a comment / URL to help the reader?

PR #1370, opened 2025-02-06
xuy-UID2-4653-remove-loki

clarkxuyang:
Tested locally with workspace


mcollins-ttd:
There's also an environment variable `LOKI_HOSTNAME` set in `scripts/azure-cc/Dockerfile`

PR #1405, opened 2025-02-13
Operator Runtime Config Rotating Store Verticle

BehnamMozafari:
- Add RotatingRuntimeConfigStore to load operator runtime config using UIDCoreClient 
- Publishes runtime config to event bus
- Change Main to inject event-bus config retriever into ConfigService
- Update config files - add `runtime_config_path` and `runtime_config_refresh_ms`

### DIFF
--- a/src/main/java/com/uid2/operator/Main.java
+++ b/src/main/java/com/uid2/operator/Main.java
@@ -273,7 +273,7 @@ public class Main {
         }
     }
 
-    private Future<IConfigService> initialiseConfigService(ConfigRetriever dynamicConfigRetriever) throws Exception {
+    private Future<IConfigService> initialiseConfigService(ConfigRetriever dynamicConfigRetriever) {
         Promise<IConfigService> promise = Promise.promise();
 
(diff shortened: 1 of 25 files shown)


src/main/java/com/uid2/operator/Main.java
```diff
@@ -337,9 +335,17 @@ private void run() throws Exception {
...
+                vertx,
+                new JsonObject()
+                        .put("type", "event-bus")
+                        .put("config", new JsonObject()
+                                .put("address", OperatorRuntimeConfigEventBus))
+        );
 
+        this.createStoreVerticles()
+                .compose(v -> this.initialiseConfigService(dynamicConfigRetriever))
+                .compose(configService -> {
```
BehnamMozafari, line 348: Reason for this change ->createStoreVerticles method needs to be before initialiseConfigService so that the runtime config is retrieved from core and published to event-bus. dynamicConfigRetriever needs to listen to runtime config address on event bus before publishing config otherwise the message will be discarded.

PR #1412, opened 2025-02-15
Abu UID2 4880 fix e2e

abuabraham-ttd:


scripts/azure-cc/azureEntryPoint.py
```diff
@@ -150,8 +146,8 @@ def _validate_auxiliaries(self):
     def run_compute(self) -> None:
         """Main execution flow for confidential compute."""
         self.__check_env_variables()
-        self.__create_final_config()
         self._set_confidential_config()
+        self.__create_final_config()
```
cYKatherine, line 150: This should stay as before

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -50,17 +49,14 @@ def __create_final_config(self):
         except IOError as e:
             logging.error(f"Failed to create {AzureEntryPoint.FINAL_CONFIG} with error: {e}")
             sys.exit(1)
-
-        CORE_BASE_URL = os.getenv("CORE_BASE_URL")
-        OPTOUT_BASE_URL = os.getenv("OPTOUT_BASE_URL")
         
-        if CORE_BASE_URL and OPTOUT_BASE_URL and AzureEntryPoint.env_name != 'prod':
-            logging.info(f"-- replacing URLs by {CORE_BASE_URL} and {OPTOUT_BASE_URL}")
+        if self.configs["core_base_url"] and self.configs["optout_base_url"] and AzureEntryPoint.env_name != 'prod':
```
cYKatherine, line 53: Why are we removing the 
```
CORE_BASE_URL = os.getenv("CORE_BASE_URL")
OPTOUT_BASE_URL = os.getenv("OPTOUT_BASE_URL")
```
How do we pass in core/optout url using environment variables now 🤔

.github/workflows/run-e2e-tests-on-operator.yaml
```diff
@@ -106,7 +106,7 @@ on:
 jobs:
   e2e-test:
     name: E2E Test
-    uses: IABTechLab/uid2-shared-actions/.github/workflows/shared-run-e2e-tests.yaml@v3
+    uses: IABTechLab/uid2-shared-actions/.github/workflows/shared-run-e2e-tests.yaml@kcc-fix-e2e
```
cYKatherine, line 109: Adding this because we need to pass in the `skipValidations` params when starting up the azure operator, see: https://github.com/IABTechLab/uid2-shared-actions/pull/149/files

scripts/azure-cc/azureEntryPoint.py
abuabraham-ttd: Good catch! yes, I removed __set_base_urls method, we are reading it from env variables anyway.

scripts/aws/ec2.py
```diff
@@ -146,8 +146,8 @@ def __run_socks_proxy(self) -> None:
         """
         Starts the SOCKS proxy service.
         """
-        command = ["sockd", "-D"]
-        self.run_command(command)
+        command = ["sockd"]
```
clarkxuyang, line 149: there is a difference between -D and separate_process= True, as -D creates a true daemon but python subprocess still attached to the main process, and will be killed when main python process dies. Here we relies on the nitro command to keep this sockd live.

scripts/confidential_compute.py
```diff
@@ -144,7 +144,7 @@ def run_command(command, separate_process=False):
             if separate_process:
                 subprocess.Popen(command, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
             else:
-                subprocess.run(command,check=True)
+                subprocess.run(command,check=True,text=True,capture_output=False)
```
clarkxuyang, line 147: why set capture_output = false ?

scripts/azure-cc/azureEntryPoint.py
clarkxuyang: CORE_BASE_URL and OPTOUT_BASE_URL might not be set for azure, in that case, the verification script will fail.

scripts/azure-cc/azureEntryPoint.py
```diff
@@ -95,11 +84,9 @@ def _set_confidential_config(self, secret_identifier: str = None):
         self.configs["skip_validations"] = os.getenv("SKIP_VALIDATIONS", "false").lower() == "true"
         self.configs["debug_mode"] = os.getenv("DEBUG_MODE", "false").lower() == "true"
         self.configs["environment"] = AzureEntryPoint.env_name
-
-        # set self.configs["operator_key"]
+        self.configs["core_base_url"] = os.getenv("CORE_BASE_URL")
```
clarkxuyang, line 87: CORE_BASE_URL and OPTOUT_BASE_URL might not be set for azure, in that case, the verification script will fail.

abuabraham-ttd: yes, this is to handle that. Earlier implementation was making final config and validating. Hence failed before validations

scripts/azure-cc/conf/integ-uid2-config.json
```diff
@@ -1,20 +1,20 @@
...
-  "keysets_metadata_path": "https://core-integ.uidapi.com/key/keyset/refresh",
-  "keyset_keys_metadata_path": "https://core-integ.uidapi.com/key/keyset-keys/refresh",
-  "client_side_keypairs_metadata_path": "https://core-integ.uidapi.com/client_side_keypairs/refresh",
-  "salts_metadata_path": "https://core-integ.uidapi.com/salt/refresh",
-  "services_metadata_path": "https://core-integ.uidapi.com/services/refresh",
-  "service_links_metadata_path": "https://core-integ.uidapi.com/service_links/refresh",
-  "optout_metadata_path": "https://optout-integ.uidapi.com/optout/refresh",
-  "core_attest_url": "https://core-integ.uidapi.com/attest",
-  "optout_api_uri": "https://optout-integ.uidapi.com/optout/replicate",
+  "sites_metadata_path": "https://core.uidapi.com/sites/refresh",
```
abuabraham-ttd, line 2: Made few modifications here to match with GCP implementation. Also, our long term goal is to remove these url paths added in configs; to keep it simple. 

@clarkxuyang to review and approve before merge.

scripts/azure-cc/conf/integ-uid2-config.json
```diff
@@ -1,20 +1,20 @@
...
+  "keysets_metadata_path": "https://core.uidapi.com/key/keyset/refresh",
+  "keyset_keys_metadata_path": "https://core.uidapi.com/key/keyset-keys/refresh",
+  "client_side_keypairs_metadata_path": "https://core.uidapi.com/client_side_keypairs/refresh",
+  "salts_metadata_path": "https://core.uidapi.com/salt/refresh",
+  "services_metadata_path": "https://core.uidapi.com/services/refresh",
+  "service_links_metadata_path": "https://core.uidapi.com/service_links/refresh",
+  "optout_metadata_path": "https://optout.uidapi.com/optout/refresh",
+  "core_attest_url": "https://core.uidapi.com/attest",
+  "optout_api_uri": "https://optout.uidapi.com/optout/replicate",
   "optout_s3_folder": "uid-optout-integ/",
```
abuabraham-ttd, line 13: `optout_s3_folder` is the only difference between integ and prod. There is an epic to use optout url returned by core and remove need for `optout_s3_folder` 

cc: @clarkxuyang

scripts/azure-cc/conf/integ-uid2-config.json
clarkxuyang: Chat with Abu, the integ and prod configuration is a "template" file. It is not a good design as the integ or prod file should has the final value for integ and prod to avoid integ accidentally pointing to prod end point. It is expected that this logic to be updated in the configuration optimization task.

PR #1558, opened 2025-03-12
Re-adding encryption

abuabraham-ttd:
Re-adding whatever was removed in https://github.com/IABTechLab/uid2-operator/pull/1265/files


cody-constine-ttd (approved):

src/main/java/com/uid2/operator/Main.java
```diff
@@ -136,25 +138,47 @@ public Main(Vertx vertx, JsonObject config) throws Exception {
...
+        if (this.encryptedCloudFilesEnabled) {
+            String cloudEncryptionKeyMdPath = this.config.getString(Const.Config.CloudEncryptionKeysMetadataPathProp);
+            this.cloudEncryptionKeyProvider = new RotatingCloudEncryptionKeyApiProvider(fsStores,
+                    new GlobalScope(new CloudPath(cloudEncryptionKeyMdPath)));
+
+            String keypairMdPath = this.config.getString(Const.Config.ClientSideKeypairsMetadataPathProp);
+            this.clientSideKeypairProvider = new RotatingClientSideKeypairStore(fsStores,
+                    new GlobalScope(new CloudPath(keypairMdPath)), cloudEncryptionKeyProvider);
+            String clientsMdPath = this.config.getString(Const.Config.ClientsMetadataPathProp);
+            this.clientKeyProvider = new RotatingClientKeyProvider(fsStores, new GlobalScope(new CloudPath(clientsMdPath)),
```
abuabraham-ttd, line 151: Thanks a lot for doing this. 

Mind if I update this to something like.

```
private RotatingClientKeyProvider createClientKeyProvider(boolean encrypted) {
    String path = config.getString(Const.Config.ClientsMetadataPathProp);
    return encrypted
        ? new RotatingClientKeyProvider(fsStores, new GlobalScope(new CloudPath(path)), cloudEncryptionKeyProvider)
        : new RotatingClientKeyProvider(fsStores, new GlobalScope(new CloudPath(path)));
}
```

We tend to duplicate a lot.

conf/default-config.json
```diff
@@ -30,6 +30,9 @@
   "salts_metadata_path": "salts/metadata.json",
   "services_metadata_path": "services/metadata.json",
   "service_links_metadata_path": "service_links/metadata.json",
+  "cloud_encryption_keys_metadata_path": "cloud_encryption_keys/metadata.json",
+  "encrypted_files": true,
```
vishalegbert-ttd, line 34: I think this should default to false to prevent it being accidentally enabled in integ/prod for now

src/main/java/com/uid2/operator/Main.java
vishalegbert-ttd: I think the way it is now will make it easier to clean up the switch

After the switch is removed, will your function make sense?

conf/default-config.json
vishalegbert-ttd: This config file packaged into the docker image as the default settings, so if we don't explicitly set this config to false in k8s manifests, it will be true

abuabraham-ttd: good point! 👍

src/main/java/com/uid2/operator/Main.java
abuabraham-ttd: > After the switch is removed 

If we do that, it is fine

PR #1563, opened 2025-03-13
Add Azure AKS workflow

mcollins-ttd:
- Modify Azure workflow to create AKS deployment artifacts.
- Use same Docker image for Azure Confidential Containers (CC) and AKS with Virtual Nodes on Azure Container Instances.
- Delete the steps to create a release from the Azure workflow, as functionality is not needed.


.github/workflows/publish-azure-cc-enclave-docker.yaml
```diff
@@ -159,97 +162,98 @@ jobs:
...
+    name: Create Azure CC artifacts
+    runs-on: ubuntu-latest
+    permissions: {}
+    needs: buildImage
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Azure CLI
+        uses: ./.github/actions/install_az_cli
```
cYKatherine, line 175: nit: `IABTechLab/uid2-operator/.github/actions/install_az_cli` to keep the repo consistent

.github/workflows/publish-azure-cc-enclave-docker.yaml
```diff
@@ -159,97 +162,98 @@ jobs:
...
+    name: Create Azure AKS artifacts
+    runs-on: ubuntu-latest
+    permissions: {}
+    needs: buildImage
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Azure CLI
+        uses: ./.github/actions/install_az_cli
```
cYKatherine, line 223: nit: IABTechLab/uid2-operator/.github/actions/install_az_cli to keep the repo consistent

scripts/azure-aks/deployment/generate.py
```diff
@@ -0,0 +1,20 @@
+import sys
```
cYKatherine, line 1: This is the same file https://github.com/IABTechLab/uid2-operator/blob/main/scripts/azure-cc/deployment/generate.py 🤔 Maybe refer to that instead of creating a same duplicated file.

.github/workflows/publish-azure-cc-enclave-docker.yaml
mcollins-ttd: Isn't the downside to that, that it will run the action as it is on the default branch? Or if there's no such action on the default branch - like in this case - you'll get an error?

mcollins-ttd: Also curious why our workflows call other workflows using `./.github/workflows/...`, but out workflows calls actions using `IABTechLab/uid2-operator/.github/actions/...`.

cYKatherine (approved):

PR #1598, opened 2025-03-24
Fix debug logs

abuabraham-ttd:
/usr/sbin/syslog-ng --verbose

is only needed when operator has debug mode enabled

### DIFF
--- a/scripts/aws/entrypoint.sh
+++ b/scripts/aws/entrypoint.sh
@@ -21,8 +21,6 @@ ifconfig lo 127.0.0.1
 echo "Starting vsock proxy..."
 /app/vsockpx --config /app/proxies.nitro.yaml --daemon --workers $(( ( $(nproc) + 3 ) / 4 )) --log-level 3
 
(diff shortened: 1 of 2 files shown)


scripts/aws/entrypoint.sh
```diff
@@ -85,6 +83,7 @@ LOGBACK_CONF="./conf/logback.xml"
 
 if [[ "$DEBUG_MODE" == "true" ]]; then
   LOGBACK_CONF="./conf/logback-debug.xml"
+  /usr/sbin/syslog-ng --verbose
 fi
```
abuabraham-ttd, line 87: reverting to how it was before. 

https://github.com/IABTechLab/uid2-operator/commit/62421ce7d26371c8ada45d103736129c31a6d496#diff-0baf5acc27c3eef7d6ea3c338f48ebcf776dd261d3b847e7e2e8b71c4038efb8L42

clarkxuyang (approved):
LGTM

PR #1636, opened 2025-04-01
UID2-4781 Add RuntimeConfig

cYKatherine:


src/main/java/com/uid2/operator/model/RuntimeConfig.java
```diff
@@ -0,0 +1,68 @@
+package com.uid2.operator.model;
```
mcollins-ttd, line 1: Not sure if model is the right package for this, maybe we should have a new package for runtime config classes

src/main/java/com/uid2/operator/model/RuntimeConfig.java
```diff
@@ -0,0 +1,68 @@
...
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import static com.uid2.operator.service.ConfigValidatorUtil.*;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class RuntimeConfig {
+    @JsonProperty("identity_token_expires_after_seconds")
+    private Integer identityTokenExpiresAfterSeconds;
```
mcollins-ttd, line 12: Can the required properties all be `int` instead of `Integer`?

src/main/java/com/uid2/operator/model/RuntimeConfig.java
```diff
@@ -0,0 +1,68 @@
...
+    @JsonProperty("refresh_identity_token_after_seconds")
+    private Integer refreshIdentityTokenAfterSeconds;
+    @JsonProperty("sharing_token_expiry_seconds")
+    private Integer sharingTokenExpirySeconds;
+    @JsonProperty("max_bidstream_lifetime_seconds")
+    private Integer maxBidstreamLifetimeSeconds;
+    @JsonProperty("max_sharing_lifetime_seconds")
+    private Integer maxSharingLifetimeSeconds;
+
+    public Integer getIdentityTokenExpiresAfterSeconds() {
```
mcollins-ttd, line 24: Can these all return `int` instead?

src/main/java/com/uid2/operator/service/ConfigService.java
```diff
@@ -41,31 +42,20 @@ public static Future<ConfigService> create(ConfigRetriever configRetriever) {
     }
 
     @Override
-    public JsonObject getConfig() {
-        return configRetriever.getCachedConfig();
+    public RuntimeConfig getConfig() {
+        return configRetriever.getCachedConfig().mapTo(RuntimeConfig.class);
```
mcollins-ttd, line 46: TODO: We don't want to map every time this method is called.

PR #1641, opened 2025-04-01
Remote config: Use rotating store verticle and map to POJO

cYKatherine:
- Periodically retrieve runtime config using a rotating store verticle, as core endpoint now requires attestation
- Update runtime config format to include an incrementing version number, as required by rotating store verticle
- Map runtime config JSON to POJO (`RuntimeConfig`)
- Validate runtime config value in `RuntimeConfig` constructor so consumers don't have to remember to call `isValid`
- Use builder pattern for runtime config. Makes tests cleaner. Can be replaced with Lombok in the future
- Log runtime config on retrieval to help with debugging

Tested scenarios:
- Run locally with feature flag OFF
  - Operator loads runtime config values from bootstrap config
- Run locally with feature flag ON, `storage_mock: false` and `runtime_config_metadata_path` pointing to core
  - Operator loads runtime config values from core API endpoint 
- Run locally with feature flag ON, `storage_mock: true` and `runtime_config_metadata_path` pointing to new `metadata.json` file
  - Operator loads runtime config values from local `metadata.json` file


src/main/java/com/uid2/operator/Main.java
```diff
@@ -385,6 +366,13 @@ private Future<Void> createStoreVerticles() throws Exception {
             siteProvider.getMetadata();
             clientSideKeypairProvider.getMetadata();
         }
+        if (useRemoteConfig) {
+            configStore.getMetadata();
```
cYKatherine, line 370: Is this required? Or should we just do `configStore.loadContent`? Don't see much meaning adding this `getMetadata()` to the `IConfigStore` interface.

src/main/java/com/uid2/operator/store/RuntimeConfig.java
```diff
@@ -0,0 +1,87 @@
...
+        } else {
+            return sharingTokenExpirySeconds;
+        }
+    }
+
+    public Integer getSharingTokenExpirySeconds() {
+        return sharingTokenExpirySeconds;
+    }
+
+    // @JsonIgnore is needed to exclude 'valid' field from JSON conversion via JsonObject.mapFrom()."
```
cYKatherine, line 58: Do we still need the `@JsonIgnore` 🤔

src/main/java/com/uid2/operator/Main.java
mcollins-ttd: I can't see why it would be required.

Ideally `IConfigStore` would just gave a single method `getConfig`.

src/test/java/com/uid2/operator/RuntimeConfigStoreTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @AfterEach
+    public void teardown() throws Exception {
+        mocks.close();
+    }
+
+    @Test
+    public void loadRuntimeConfigSingleVersion(Vertx vertx) throws Exception {
+        final JsonObject metadataJson = new JsonObject();
+        {
```
mcollins-ttd, line 56: Why the `{}`s?

src/test/java/com/uid2/operator/RuntimeConfigStoreTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+        when(cloudStorage.download("metadata"))
+                .thenReturn(new ByteArrayInputStream(metadataJson.toString().getBytes(StandardCharsets.US_ASCII)));
+
+        RuntimeConfigStore runtimeConfigStore = new RuntimeConfigStore(vertx, cloudStorage, "metadata");
+
+        final JsonObject loadedMetadata = runtimeConfigStore.getMetadata();
+        runtimeConfigStore.loadContent(loadedMetadata);
+        assertEquals(2, runtimeConfigStore.getVersion(loadedMetadata));
+        assertEquals(this.runtimeConfig, runtimeConfigStore.getConfig());
```
mcollins-ttd, line 69: Instead of overriding `equals` we can use AssertJ:

```java
        assertThat(runtimeConfigStore.getConfig())
                .usingRecursiveComparison()
                .isEqualTo(this.runtimeConfig);
```

- It has better error messages that tell us which fields differ
- We don't have to write and maintain `equals`
- We don't have to write and maintain `hashCode`

src/main/java/com/uid2/operator/store/RuntimeConfig.java
```diff
@@ -0,0 +1,87 @@
...
+    @JsonIgnore
+    public boolean isValid() {
```
mcollins-ttd, line 60: Can we do validation in a constructor so that clients don't have to remember to call `isValid` every time?

mcollins-ttd (approved):

PR #1687, opened 2025-04-10
sch-UID2-4635 Added logging when cstg key matches an api key

sophia-chen-ttd:


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -341,6 +341,14 @@ private Set<String> getAppNames(ClientSideKeypair keypair) {
         return site.getAppNames();
     }
 
+    private void checkInvalidApiKey(String key) {
```
aulme, line 344: Probably can have a better name, how about something like `logIfApiKey` - you actually have that as the first line comment.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -3673,6 +3679,7 @@ void cstgBadPublicKey(Vertx vertx, VertxTestContext testContext) throws NoSuchAl
                 400,
                 testContext,
                 respJson -> {
+                    Assertions.assertEquals(publicKey.equals(clientKey), logWatcher.list.stream().anyMatch(l -> l.getFormattedMessage().contains("Client side key is an api key with api_key_id=key-id for site_id=1")));
```
aulme, line 3682: Great that there's a test! It ends up being a little convoluted because of the good/bad case parametrization though. I think it will be simpler if you move this out into smaller separate tests that just assert that logging happens when there is a match and doesn't when there isn't.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -3719,6 +3731,7 @@ void cstgBadSubscriptionId(Vertx vertx, VertxTestContext testContext) throws NoS
                 400,
                 testContext,
                 respJson -> {
+                    Assertions.assertEquals(subscriptionId.equals(clientKey), logWatcher.list.stream().anyMatch(l -> l.getFormattedMessage().contains("Client side key is an api key with api_key_id=key-id for site_id=1")));
```
aulme, line 3734: Same as above

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
sophia-chen-ttd: They're both bad cases in the tests, one is a cstg key that doesn't exist and the other is a client-key. I don't think I can move them out into smaller separate tests because the only 'public' method is the endpoint itself and everything in the test is needed to send the actual request.

vishalegbert-ttd: I think what Aleks means is that it's unclear which key is causing the log to be written

vishalegbert-ttd (approved):

PR #1752, opened 2025-05-19
sch-UID2-5484 v3 identity map operator logic

sophia-chen-ttd:


src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -218,20 +221,48 @@ private byte[] getFirstLevelHash(byte[] identityHash, Instant asOf) {
...
+
+    private byte[] getAdvertisingId(UserIdentity firstLevelHashIdentity, String salt) {
+        return rawUidV3Enabled
+                ? TokenUtils.getAdvertisingIdV3(firstLevelHashIdentity.identityScope, firstLevelHashIdentity.identityType, firstLevelHashIdentity.id, salt)
+                : TokenUtils.getAdvertisingIdV2(firstLevelHashIdentity.id, salt);
+    }
+
+    private byte[] getPreviousAdvertisingId(UserIdentity firstLevelHashIdentity, SaltEntry rotatingSalt, Instant asOf) {
+        long age = asOf.toEpochMilli() - rotatingSalt.lastUpdated();
+        if ( age / DAY_IN_MS < 90) {
```
vishalegbert-ttd, line 245: Should this be less than 90 or less than or equal to 90? 

cc @aulme

src/main/java/com/uid2/operator/model/MappedIdentity.java
```diff
@@ -1,13 +1,17 @@
 package com.uid2.operator.model;
 
 public class MappedIdentity {
-    public static MappedIdentity LogoutIdentity = new MappedIdentity(new byte[33], "");
+    public static MappedIdentity LogoutIdentity = new MappedIdentity(new byte[33], "", null, null);
     public final byte[] advertisingId;
     public final String bucketId;
+    public final byte[] previousId;
```
vishalegbert-ttd, line 7: previousAdvertisingId? 

Otherwise at first glance it seems like an unrelated property

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -218,20 +221,48 @@ private byte[] getFirstLevelHash(byte[] identityHash, Instant asOf) {
...
+    }
+
+    private byte[] getAdvertisingId(UserIdentity firstLevelHashIdentity, String salt) {
+        return rawUidV3Enabled
+                ? TokenUtils.getAdvertisingIdV3(firstLevelHashIdentity.identityScope, firstLevelHashIdentity.identityType, firstLevelHashIdentity.id, salt)
+                : TokenUtils.getAdvertisingIdV2(firstLevelHashIdentity.id, salt);
+    }
+
+    private byte[] getPreviousAdvertisingId(UserIdentity firstLevelHashIdentity, SaltEntry rotatingSalt, Instant asOf) {
+        long age = asOf.toEpochMilli() - rotatingSalt.lastUpdated();
```
gmsdelmundo, line 244: `lastUpdated` shouldn't have an issue if it's set to UTC midnight? i.e. the truncation of the trailing zeroes vs `asOf.toEpochMilli()`

src/main/java/com/uid2/operator/service/UIDOperatorService.java
gmsdelmundo, line 245: Extra space

src/main/java/com/uid2/operator/service/UIDOperatorService.java
```diff
@@ -218,20 +221,48 @@ private byte[] getFirstLevelHash(byte[] identityHash, Instant asOf) {
...
+            }
+            return getAdvertisingId(firstLevelHashIdentity, rotatingSalt.previousSalt());
+        }
+        return null;
+    }
+
+    private long getRefreshFrom(SaltEntry rotatingSalt, Instant asOf) {
+        Long refreshFrom = rotatingSalt.refreshFrom();
+        if (refreshFrom == null || refreshFrom < asOf.toEpochMilli()) {
+            return asOf.truncatedTo(DAYS).plus(1, DAYS).toEpochMilli();
```
gmsdelmundo, line 257: I think we can go with `getEpochSecond()` for `refreshFrom`'s value in the API

There shouldn't be an issue with the epoch precision in this case since it's a new API

src/main/java/com/uid2/operator/model/MappedIdentity.java
```diff
@@ -1,13 +1,17 @@
...
+    public final Long refreshFrom;
```
gmsdelmundo, line 8: With these new fields, will it affect the response payload of `/v2/identity/map`?

gmsdelmundo: Oh, probably not since we manually construct the response in `handleIdentityMapCommon()`, but we might need to take that into consideration when implementing the actual v3 API

src/test/java/com/uid2/operator/UIDOperatorServiceTest.java
```diff
@@ -861,4 +884,107 @@ void testExpiredSaltsNotifiesShutdownHandler(TestIdentityInputType type, String
...
+        var expectedPreviousUID = UIDOperatorVerticleTest.getRawUid(IdentityType.Email, email, FIRST_LEVEL_SALT, salt.previousSalt(), IdentityScope.UID2, uid2Config.getBoolean(IdentityV3Prop));
+        assertArrayEquals(expectedCurrentUID, mappedIdentity.advertisingId);
+        assertArrayEquals(expectedPreviousUID, mappedIdentity.previousId);
+    }
+
+    @Test
+    void testMappedIdentityWithOutdatedPreviousSaltReturnsNoPreviousUid() {
+        var saltSnapshot = setUpMockSalts();
+
+        long lastUpdatedOver90DaysAgo = this.now.minus(91, DAYS).toEpochMilli();
```
gmsdelmundo, line 914: Since the condition is set to `< 90`, maybe let's set this to be exactly 90 so we can test the boundary of our condition?

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -656,7 +656,7 @@ private byte[] getAdvertisingIdFromIdentity(IdentityType identityType, String id
         return getRawUid(identityType, identityString, firstLevelSalt, rotatingSalt, getIdentityScope(), useRawUidV3());
     }
 
-    private static byte[] getRawUid(IdentityType identityType, String identityString, String firstLevelSalt, String rotatingSalt, IdentityScope identityScope, boolean useRawUidV3) {
+    public static byte[] getRawUid(IdentityType identityType, String identityString, String firstLevelSalt, String rotatingSalt, IdentityScope identityScope, boolean useRawUidV3) {
```
gmsdelmundo, line 659: Maybe not for this PR, but I think we should move the different `getRawUid()` methods to a dedicated test util class

src/test/java/com/uid2/operator/UIDOperatorServiceTest.java
```diff
@@ -861,4 +884,107 @@ void testExpiredSaltsNotifiesShutdownHandler(TestIdentityInputType type, String
...
+        assertArrayEquals(null, mappedIdentity.previousId);
+    }
+
+    @ParameterizedTest
+    @ValueSource(strings = {"0", "1"})
+    void testMappedIdentityWithValidRefreshFrom(int refreshFromDays) {
+        var saltSnapshot = setUpMockSalts();
+
+        long lastUpdated = this.now.minus(30, DAYS).toEpochMilli();
+        long refreshFrom = this.now.plus(refreshFromDays, DAYS).toEpochMilli();
```
gmsdelmundo, line 957: If last updated was 30 days ago, is it valid for `refreshFrom` to be 31 days? Shouldn't it be multiples of 30 days since `lastUpdated`?

src/main/java/com/uid2/operator/service/UIDOperatorService.java
sophia-chen-ttd: Yep although I think it's better to do that on the api side when we are returning the response so that we keep eveything consistent as millis internally.

src/main/java/com/uid2/operator/model/MappedIdentity.java
sophia-chen-ttd: No, the old apis only access the fields that they need and build the response using those fields

src/main/java/com/uid2/operator/service/UIDOperatorService.java
sophia-chen-ttd: lastUpdated is in millis so there won't be any truncation

src/main/java/com/uid2/operator/service/UIDOperatorService.java
aulme: Either we should move this bit out into shared or at least leave a comment in both places that they need to be changed in tandem.

I think this should stay <90, the 1ms doesn't matter and we should keep this in sync with Admin where it's <90. In Admin if we do <=90 we'll end up with 91 days since we'll retain the salt on the 90th rotation, only removing it on the 91st. 
If in doubt add unit tests to verify that 1m before 90 days it's shown and 1ms after it's hidden.

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
aulme: Yeah, probably should be a builder, it's used all over. Too much for this PR though.

PR #1754, opened 2025-05-20
sch-UID2-5484 v3 identity map api

sophia-chen-ttd:


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1658,6 +1717,63 @@ private InputUtil.InputVal[] getIdentityMapV2Input(RoutingContext rc) {
                 getInputList.get();
     }
 
+    private void handleIdentityMapV3(RoutingContext rc) {
+        try {
+            final Map<String, InputUtil.InputVal[]> inputList = getIdentityMapMixedInput(rc);
+            if (inputList == null) {
```
gmsdelmundo, line 1723: If all fields are empty arrays, is that also considered an invalid state?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1960,37 +2076,45 @@ private InputUtil.InputVal[] createInputListV1(JsonArray a, IdentityType identit
...
+
+    private InputUtil.InputVal[] parseIdentitiesInput(JsonArray identities, IdentityType identityType, InputUtil.IdentityInputType inputType, RoutingContext rc) {
+        if (identities == null || identities.isEmpty()) {
+            return new InputUtil.InputVal[0];
+        }
+        final int size = identities.size();
+        final InputUtil.InputVal[] normalizedIdentities = new InputUtil.InputVal[size];
+
+        for (int i = 0; i < size; i++) {
+            JsonObject identity = JsonParseUtils.parseObjectFromArray(identities, i, rc);
```
gmsdelmundo, line 2107: Is it possible for this to return `null` outside of the exception block in `parseObjectFromArray`? Say, one of the objects in the array itself is `null`, should we go ahead and just return `null` for the whole identity type?

I'm thinking it might be better to just skip that element and continue processing the rest, and we just treat the `null` element as an invalid input

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
gmsdelmundo: e.g. when doing checks like `if (normalizedEmails == null)`, should we check `normalizedEmails.length == 0` as well?

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1658,6 +1717,63 @@ private InputUtil.InputVal[] getIdentityMapV2Input(RoutingContext rc) {
...
+        }
+
+        final JsonArray phoneHashes = this.phoneSupport ? JsonParseUtils.parseArray(obj,"phone_hash", rc) : null;
+        if (phoneHashes != null && !phoneHashes.isEmpty()) {
+            InputUtil.InputVal[] normalizedPhoneHashes = parseIdentitiesInput(phoneHashes, IdentityType.Phone, InputUtil.IdentityInputType.Hash, rc);
+            if (normalizedPhoneHashes == null) { return null; }
+            normalizedIdentities.put("phone_hash", normalizedPhoneHashes);
+        }
+
+        if (emails == null && emailHashes == null && phones == null && phoneHashes == null) {
```
gmsdelmundo, line 1770: I think currently, an identity type can be null in the following conditions:
1. The identity type itself was not specified in the input
2. One of the elements in the array are null

I think 1 is valid, but for 2, I think we should continue processing it but treating the element as an invalid identifier

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1658,6 +1717,63 @@ private InputUtil.InputVal[] getIdentityMapV2Input(RoutingContext rc) {
...
+
+            final JsonObject resp = handleIdentityMapV3(rc, inputList);
+            ResponseUtil.SuccessV2(rc, resp);
+        } catch (Exception e) {
+            ResponseUtil.LogErrorAndSendResponse(ResponseStatus.UnknownError, 500, rc, "Unknown error while mapping identity v3", e);
+        }
+    }
+
+    private Map<String, InputUtil.InputVal[]> getIdentityMapMixedInput(RoutingContext rc) {
+        final JsonObject obj = (JsonObject) rc.data().get("request");
```
gmsdelmundo, line 1738: If "request" is null, I think this might throw a NPE

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -112,7 +113,7 @@ public class UIDOperatorVerticle extends AbstractVerticle {
 
     private final Map<String, DistributionSummary> optOutStatusCounters = new HashMap<>();
     private final IdentityScope identityScope;
-    private final V2PayloadHandler v2PayloadHandler;
+    private final V2PayloadHandler encryptedPayloadHandler;
```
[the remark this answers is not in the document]
sophia-chen-ttd: I think we can do this in a separate ticket because this means we'll probably need to change all the other V2 named classes as well

src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1579,6 +1591,49 @@ private JsonObject handleIdentityMapCommon(RoutingContext rc, InputUtil.InputVal
...
+                final JsonObject resp = new JsonObject();
+                if (rawId != null && rawId.isValid()) {
+                    final MappedIdentity mappedId = idService.mapIdentity(
+                            new MapRequest(
+                                    rawId.toUserIdentity(this.identityScope, 0, now),
+                                    OptoutCheckPolicy.respectOptOut(),
+                                    now));
+
+                    if (mappedId.isOptedOut()) {
+                        resp.put("e", "OPTOUT");
```
gmsdelmundo, line 1616: Thoughts on making this an enum?

sophia-chen-ttd: Good idea

PR #1784, opened 2025-06-04
Add service instance ID and UID trace ID to core and opt out requests

vishalegbert-ttd:
Changes in this PR and [Shared PR](https://github.com/IABTechLab/uid2-shared/pull/457/files):
- Added instance ID configs for local Operators
- Initialised the UidInstanceIdProvider and passed it to where it's needed
- Attached UID-Instance-Id header to all requests made to Core and Opt Out
- Forwarded UID-Trace-Id header for logout requests

Tested locally in dev workspace:

- UID instance ID is being populated in headers in all requests to Core
- Core is receiving the instance ID and adding it to audit logs

Screenshot 2025-06-05 at 9 54 17 pm

### DIFF
--- a/conf/local-e2e-docker-private-config.json
+++ b/conf/local-e2e-docker-private-config.json
@@ -31,5 +31,6 @@
   "cloud_refresh_interval": 30,
   "salts_expired_shutdown_hours": 12,
   "operator_type": "private",
-  "enable_remote_config": false
+  "enable_remote_config": false,
+  "uid_instance_id_prefix": "local-private-operator"
 }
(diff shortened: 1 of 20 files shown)


conf/local-e2e-docker-private-config.json
```diff
@@ -31,5 +31,6 @@
   "cloud_refresh_interval": 30,
   "salts_expired_shutdown_hours": 12,
   "operator_type": "private",
-  "enable_remote_config": false
+  "enable_remote_config": false,
+  "uid_instance_id_prefix": "local-private-operator"
 }
```
abuabraham-ttd, line 36: follow a pattern, so redacting spoofed values is easier ? 

`^[a-z]{2,3}-[a-zA-Z0-9-]{1,20}-[a-zA-Z0-9-]{1,20}(-.{1,4})?$`

otherwise these values wont be shown when running in local

abuabraham-ttd: No it is fine. It is better to handle in redact than enforce this strict pattern

PR #1786, opened 2025-06-04
Add instance_id_prefix in AWS, Azure, and GCP

abuabraham-ttd:
https://github.com/IABTechLab/uid2-operator/actions/runs/15474269414 

Successful E2E. 

Changes

- Creates service_instance_id prefix and updates config
- Removed EntryPoint from class names to use class name in service_instance_id 
- Logic added in super class to ensure pattern and logging stay the same across CSP

### DIFF
--- a/scripts/gcp-oidc/gcp.py
+++ b/scripts/gcp-oidc/gcp.py
@@ -5,12 +5,27 @@ import shutil
 from typing import Dict
 import sys
 import logging
+import requests
+import re
 from google.cloud import secretmanager
 from google.auth.exceptions import DefaultCredentialsError
 from google.api_core.exceptions import PermissionDenied, NotFound
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 from confidential_compute import ConfidentialCompute, ConfidentialComputeConfig, ConfigurationMissingError, OperatorKeyNotFoundError, OperatorKeyPermissionError, ConfidentialComputeStartupError
 
+
+class AuxiliaryConfig:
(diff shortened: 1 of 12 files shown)


scripts/gcp-oidc/gcp.py
```diff
@@ -5,13 +5,28 @@
...
 from google.auth.exceptions import DefaultCredentialsError
 from google.api_core.exceptions import PermissionDenied, NotFound
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
-from confidential_compute import ConfidentialCompute, ConfidentialComputeConfig, ConfigurationMissingError, OperatorKeyNotFoundError, OperatorKeyPermissionError, ConfidentialComputeStartupError
+from confidential_compute import ConfidentialCompute, ConfigurationMissingError, OperatorKeyNotFoundError, OperatorKeyPermissionError, ConfidentialComputeStartupError
 
-class GCPEntryPoint(ConfidentialCompute):
+
+class AuxiliaryConfig:
+    GCP_METADATA: str = "169.254.169.254"
```
asloobq, line 18: what does this hardcoded IP address refer to ?

asloobq (commented):
1 question on a hardcoded IP address. Otherwise LGTM

scripts/gcp-oidc/gcp.py
abuabraham-ttd: This is the instance metadata endpoint. It is static 169.254.169.254 and local to instance (link local IP range). AWS, GCP both provides this and it returns instance metadata (region, instance-id, image used etc)

PR #1803, opened 2025-06-10
sch-UID2-5557 binary payloads

sophia-chen-ttd:
Allow binary payloads if there is a "Content-Type: application/octet-stream" header in the request for V2 and V3 endpoints.

Unit tests: Tested header on each applicable endpoint


src/main/java/com/uid2/operator/vertx/V2PayloadHandler.java
```diff
@@ -149,12 +149,18 @@ public void handleTokenRefresh(RoutingContext rc, Handler<RoutingContext> apiHan
...
 
             if (request != null) {
-                rc.response().putHeader(HttpHeaders.CONTENT_TYPE, "text/plain");
                 // Encrypt whole payload using key shared with client.
                 byte[] encryptedResp = AesGcm.encrypt(
                     respJson.encode().getBytes(StandardCharsets.UTF_8),
                     request.encryptionKey);
-                rc.response().end(Utils.toBase64String(encryptedResp));
+
+                if (rc.request().headers().contains("Content-Type", "application/octet-stream", true)) {
```
gmsdelmundo, line 157: `application/octet-stream` gets repeated a few times

We could either set the different acceptable MIME types in an enum/static strings

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -345,18 +365,23 @@ private void sendTokenRefresh(String apiVersion, Vertx vertx, VertxTestContext t
...
             refreshHttpRequest
-                    .putHeader("content-type", "text/plain")
                     .sendBuffer(Buffer.buffer(refreshToken.getBytes(StandardCharsets.UTF_8)), testContext.succeeding(response -> testContext.verify(() -> {
                         assertEquals(expectedHttpCode, response.statusCode());
 
                         if (response.statusCode() == 200 && v2RefreshDecryptSecret != null) {
-                            byte[] decrypted = AesGcm.decrypt(Utils.decodeBase64String(response.bodyAsString()), 0, Utils.decodeBase64String(v2RefreshDecryptSecret));
+                            byte[] byteResp;
+                            if (response.headers().contains("Content-Type", "application/octet-stream", true)) {
+                                byteResp =response.bodyAsBuffer().getBytes();
```
gmsdelmundo, line 380: Space after `=`

src/main/java/com/uid2/operator/util/HttpMediaType.java
```diff
@@ -0,0 +1,7 @@
+package com.uid2.operator.util;
+
+public final class HttpMediaType {
+    public static final String TEXT_PLAIN = "text/plain";
+    public static final String APPLICATION_OCTET_STREAM = "application/octet-stream";
+    public static final String APPLICATION_JSON = "application/json";
+}
```
gmsdelmundo, line 7: You might also want to add a private constructor

i.e.
```
private HttpMediaType() {
}
```

We don't need/want this class to be instantiated since it's only used to contain acceptable MIME types

gmsdelmundo: Or you could switch this to an enum, which I think is the better option

```
public enum HttpMediaType {
  TEXT_PLAIN("text/plain"),
  APPLICATION_JSON("application/json"),
  APPLICATION_OCTET_STREAM("application/octet-stream");

  private final String type;

  public HttpMediaType(String type) {
    this.type = type;
  }
}
```

src/main/java/com/uid2/operator/vertx/V2PayloadHandler.java
```diff
@@ -174,18 +181,30 @@ private void passThrough(RoutingContext rc, Handler<RoutingContext> apiHandler)
             return;
         }
         JsonObject respJson = (JsonObject) rc.data().get("response");
-        rc.response().putHeader(HttpHeaders.CONTENT_TYPE, "application/json")
+        rc.response().putHeader(HttpHeaders.CONTENT_TYPE, HttpMediaType.APPLICATION_JSON.getType())
             .end(respJson.encode());
     }
 
-    private void writeResponse(RoutingContext rc, byte[] nonce, JsonObject resp, byte[] keyBytes) {
+    public static byte[] writeResponseBody(byte[] nonce, JsonObject resp, byte[] keyBytes) {
```
vishalegbert-ttd, line 188: I think this can be renamed to encryptResponse

src/main/java/com/uid2/operator/vertx/V2PayloadHandler.java
```diff
@@ -174,18 +181,30 @@ private void passThrough(RoutingContext rc, Handler<RoutingContext> apiHandler)
...
-        rc.response().putHeader(HttpHeaders.CONTENT_TYPE, "text/plain");
-        rc.response().end(Utils.toBase64String(AesGcm.encrypt(buffer.getBytes(), keyBytes)));
+        byte[] response = buffer.getBytes();
+        return AesGcm.encrypt(response, keyBytes);
+    }
+
+    private void writeResponse(RoutingContext rc, byte[] nonce, JsonObject resp, byte[] keyBytes) {
+        var response = writeResponseBody(nonce, resp, keyBytes);
+
+        if (rc.request().headers().contains(HttpHeaders.CONTENT_TYPE, HttpMediaType.APPLICATION_OCTET_STREAM.getType(), true)) {
```
vishalegbert-ttd, line 201: This if statement logic is duplicated, keeping them consistent is pretty critical, let's extract it into its own function: writeResponseBody

src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -928,7 +963,7 @@ void identityMapNewClientNoPolicySpecified(String apiVersion, Vertx vertx, Vertx
             Assertions.assertEquals(emails.getString(0), unmappedArr.getJsonObject(0).getString("identifier"));
             Assertions.assertEquals("optout", unmappedArr.getJsonObject(0).getString("reason"));
             testContext.completeNow();
-        });
+        }, Map.of("Content-Type", contentType));
```
vishalegbert-ttd, line 966: A few raw "Content-Type" strings were missed

PR #1830, opened 2025-06-23
update main file to use logger instead of print

lizk886:


src/main/java/com/uid2/operator/Main.java
```diff
@@ -457,7 +457,7 @@ private static Vertx createVertx() {
             MBeanServer server = ManagementFactory.getPlatformMBeanServer();
             server.registerMBean(AdminApi.instance, objectName);
         } catch (InstanceAlreadyExistsException | MBeanRegistrationException | NotCompliantMBeanException | MalformedObjectNameException e) {
-            System.err.format("%s", e.getMessage());
+            LOGGER.error(e.getMessage(), e);
```
sunnywu, line 460: Should there be some keywords added e.g. `"mBean initialisation failed {}", e.getMessage() `for easier code search? and do we need to log `e `explicitly or e.getMessage() is sufficient?

lizk886: Logger.error() takes in throwable as a 2nd argument would print stack tracings.

Will update the error message as suggested!

lizk886: https://stackoverflow.com/questions/10477607/avoid-printstacktrace-use-a-logger-call-instead

PR #1870, opened 2025-07-08
Draft proposal; Support autoupdate in AWS Private operators

abuabraham-ttd:


scripts/aws/logrotate/autoupdate
```diff
@@ -0,0 +1,5 @@
+# Run the daily autoupdate job
```
abuabraham-ttd, line 1: cron job to check and update

PR #1930, opened 2025-07-28
remove v1 related conditional check code

lizk886:
Plan to squash and merge due to long commit history

### DIFF
--- a/src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
+++ b/src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
@@ -409,25 +409,6 @@ public class UIDOperatorVerticleTest {
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/test/java/com/uid2/operator/UIDOperatorVerticleTest.java
```diff
@@ -276,12 +276,6 @@ private void send(String apiVersion, Vertx vertx, String endpoint, boolean isV1G
                     handler.handle(tryParseResponse(ar.result()));
                 }
             }, additionalHeaders);
-        } else if (isV1Get) {
-            get(vertx, endpoint + (v1GetParam != null ? "?" + v1GetParam : ""), ar -> {
```
vishalegbert-ttd, line 280: Can we get rid of the get function entirely? I think it's unused now as all v2/v3 endpoints use post

PR #1980, opened 2025-09-02
sch-UID2-5966 added identity environment scope to remote config

sophia-chen-ttd:


src/main/java/com/uid2/operator/model/IdentityEnvironment.java
```diff
@@ -0,0 +1,33 @@
+package com.uid2.operator.model;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.uid2.operator.vertx.ClientInputValidationException;
+
+public enum IdentityEnvironment {
+    TEST(0), INTEG(1), PROD(2);
+
+    public final int value;
```
gmsdelmundo, line 9: This can be `private` and just add a getter

gmsdelmundo (approved):

PR #2542, opened 2026-05-14
UID2-7002 Use 'publish-major' env instead of hardcoded approver list

swibi-ttd:
Replaces `trstringer/manual-approval` steps in both `publish-all-operators.yaml` and `publish-public-operator-docker-image.yaml` with env-gated `check_major` jobs using the new `publish-major` environment.

Behaviour preserved:
  - Major release dispatched via `publish-all-operators.yaml` → 2 approval prompts (parent + leaf), same as before
  - Major denied at either gate blocks all downstream builds
  - Major dispatched directly on `publish-public-operator-docker-image.yaml` → 1 prompt (leaf)
  - Non-Major / scheduled patch → no gates

### DIFF
--- a/.github/workflows/publish-all-operators.yaml
+++ b/.github/workflows/publish-all-operators.yaml
@@ -22,8 +22,16 @@ on:
     - cron: "0 0 * * *"
 
 jobs:
+  check_major:
+    name: Check if major release
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.release_type == 'Major' && 'publish-major' || '' }}
+    steps:
+      - run: echo "Major release approved"
+
   start:
     name: Start Operator Build
+    needs: check_major
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-all-operators.yaml
```diff
@@ -22,8 +22,16 @@ on:
     - cron: "0 0 * * *"
 
 jobs:
+  check_major:
+    name: Check if major release
+    runs-on: ubuntu-latest
+    environment: ${{ inputs.release_type == 'Major' && 'publish-major' || '' }}
+    steps:
+      - run: echo "Major release approved"
```
BehnamMozafari: I think this will always print "Major release approved" even if its not a major release, maybe make this step conditional on whether its a major release?

swibi-ttd: good catch, would have been scary to see that if intention was a Patch release lol

have updated to only print that on a Major release

BehnamMozafari (approved):

PR #2578, opened 2026-06-03
chore: upgrade GitHub Actions from Node 20 to Node 24

mcollins-ttd:
Node 20 is being deprecated on GitHub Actions runners. Starting **June 16, 2026**, runners will use Node 24 by default.

This PR upgrades outdated action references in this repo to the latest released version that ships Node 24 support, pinned to its commit SHA. Each action's node version is determined by reading its `action.yml` at the pinned ref.

See: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/.github/workflows/build-uid2-ami.yaml
+++ b/.github/workflows/build-uid2-ami.yaml
@@ -38,7 +38,7 @@ jobs:
       enclave_id: ${{ steps.buildAMI.outputs.enclave_id }}
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       - name: Build UID2 Operator AMI
         id: buildAMI
@@ -78,7 +78,7 @@ jobs:
         enclave_id: ${{ steps.buildAMI.outputs.enclave_id }}
       steps:
         - name: Checkout repo
-          uses: actions/checkout@v4
+          uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
         - name: Pre-cleanup
           shell: bash
@@ -123,26 +123,26 @@ jobs:
     needs: [buildUID2, testUID2Ami, testEUIDAmi]
     steps:
       - name: Download UID2 artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: uid2_AMI_measurement
           path: ./artifacts
 
       - name: Download EUID artifacts
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
           name: euid_AMI_measurement
           path: ./artifacts
 
       - name: Delete staging artifacts
-        uses: geekyeggo/delete-artifact@b54d29a59e55046d1f7fc8226cdda507e6b9cf62 # v5
+        uses: geekyeggo/delete-artifact@176a747ab7e287e3ff4787bf8a148716375ca118 # v6.0.0
         with: 
           name: |
             uid2_AMI_measurement
             euid_AMI_measurement
 
       - name: Upload artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
           name: aws-ami-ids-${{ needs.buildUID2.outputs.version_number }}
           path: ./artifacts/
--- a/.github/workflows/publish-aws-eks-nitro-enclave-docker.yaml
+++ b/.github/workflows/publish-aws-eks-nitro-enclave-docker.yaml
@@ -37,7 +37,7 @@ jobs:
       packages: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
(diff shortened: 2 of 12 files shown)


.github/workflows/publish-gcp-oidc-enclave-docker.yaml
mcollins-ttd, line 1: The large diff in `publish-gcp-oidc-enclave-docker.yaml` includes CRLF -> LF normalization alongside the actual changes. It was the only workflow file using CRLF.

.github/workflows/build-uid2-ami.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
       enclave_id: ${{ steps.buildAMI.outputs.enclave_id }}
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
```
mcollins-ttd, line 41: **Claude Review:** This step is adjacent to a call to `./.github/actions/build_ami`, which internally still uses `actions/checkout@v4` (Node 20) and `actions/upload-artifact@v4` (Node 20). Upgrading the checkout here in the calling workflow won't silence the Node 20 deprecation warning that comes from inside the composite action. The composite action at `.github/actions/build_ami/action.yaml` also needs its action references updated to complete this fix.

.github/workflows/publish-aws-eks-nitro-enclave-docker.yaml
```diff
@@ -128,7 +128,7 @@ jobs:
           echo "Enclave ID (maybe shared by other images): " ${{ needs.buildEUIDImage.outputs.enclave_id }} >> ${{ env.ARTIFACTS_BASE_OUTPUT_DIR }}/manifests/aws-eks-euid-enclave-id-${{ needs.buildEUIDImage.outputs.image_tag }}.txt
 
       - name: Save Manifests as build artifacts
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```
mcollins-ttd, line 131: **Claude Review:** The `build_eks_docker_image` composite action (called earlier in this workflow) still pins `actions/checkout@v4`, `docker/login-action` v3, `docker/metadata-action` v5, and `docker/build-push-action` v5 — all Node 20. This composite action needs to be updated alongside the calling workflow, otherwise Node 20 deprecation warnings will continue to appear for every run. The composite also references `download_release_artifact`, which uses `actions/github-script@v7` (also Node 20).

.github/workflows/publish-aws-nitro-eif.yaml
```diff
@@ -74,7 +74,7 @@ jobs:
     needs: start
```
mcollins-ttd, line 74: **Claude Review:** The `build_aws_eif` composite action called by this workflow still uses `actions/checkout@v4` (Node 20) internally. `.github/actions/build_aws_eif/action.yaml` needs upgrading alongside the calling workflows to fully eliminate Node 20 deprecation warnings.

.github/workflows/publish-aws-nitro-eif.yaml
```diff
@@ -182,7 +182,7 @@ jobs:
       - name: Build changelog
         id: github_release
```
mcollins-ttd, line 183: **Claude Review:** `mikepenz/release-changelog-builder-action` is being upgraded from v4 directly to v6.2.2, skipping v5 which introduced breaking changes to regex transformer behavior (transformers now use `replace` by default instead of `match`). The current `configurationJson` here only uses `template` and `pr_template` (no custom transformers), so this specific usage should be unaffected — but worth flagging so the team knows to check if custom transformer config is ever added in future.

.github/workflows/publish-azure-cc-enclave-docker.yaml
```diff
@@ -88,7 +88,7 @@ jobs:
           github_token: ${{ github.ref_protected && secrets.GH_MERGE_TOKEN || '' }}
```
mcollins-ttd, line 88: **Claude Review:** The `update_operator_version` composite action called by this workflow still contains `actions/checkout@v4` steps and uses `actions/github-script@v7` (via `download_release_artifact`), both Node 20. The composite at `.github/actions/update_operator_version/action.yaml` needs to be updated to complete the Node 24 migration for workflows that call it.

.github/workflows/publish-gcp-oidc-enclave-docker.yaml
```diff
@@ -1,245 +1,245 @@
...
+          load: true
+          tags: ${{ steps.meta-all.outputs.tags }}
+          labels: ${{ steps.meta-all.outputs.labels }}
+          build-args: |
+            JAR_VERSION=${{ steps.update_version.outputs.new_version }}
+            IMAGE_VERSION=${{ steps.update_version.outputs.new_version }}
+            BUILD_TARGET=${{ env.ENCLAVE_PROTOCOL }}
+
+      - name: Vulnerability Scan
+        uses: IABTechLab/uid2-shared-actions/actions/vulnerability_scan@v3
```
mcollins-ttd, line 165: **Claude Review:** `uid2-shared-actions/actions/vulnerability_scan@v3` is the only action in this PR still using a mutable version tag instead of a pinned commit SHA. Every other action has been hardened to a SHA. Worth either pinning this to a commit SHA (for supply-chain security parity with the rest of the PR) or opening a follow-up ticket to track it — especially if that action internally uses a Node 20 runner.

sophia-chen-ttd (approved):

PR #2590, opened 2026-06-09
chore: upgrade GitHub Actions from Node 20 to Node 24

mcollins-ttd:
Node 20 is being deprecated on GitHub Actions runners. Starting **June 16, 2026**, runners will use Node 24 by default.

This PR upgrades outdated action references in this repo to the latest released version that ships Node 24 support, pinned to its commit SHA. Each action's node version is determined by reading its `action.yml` at the pinned ref.

See: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/.github/actions/build_ami/action.yaml
+++ b/.github/actions/build_ami/action.yaml
@@ -71,7 +71,7 @@ runs:
 
     - name: Get EIF for Run ${{ inputs.operator_run_number }}
       id: get_eif_for_run
-      uses: dawidd6/action-download-artifact@bf251b5aa9c2f7eeb574a96ee720e24f801b7c11 # v6
+      uses: dawidd6/action-download-artifact@b6e2e70617bc3265edd6dab6c906732b2f1ae151 # v21
       if: ${{ inputs.operator_release == '' }}
       with:
         name: 'aws-${{ inputs.identity_scope }}-deployment-files-.*'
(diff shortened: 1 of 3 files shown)


.github/actions/build_ami/action.yaml
```diff
@@ -71,7 +71,7 @@ runs:
 
     - name: Get EIF for Run ${{ inputs.operator_run_number }}
       id: get_eif_for_run
-      uses: dawidd6/action-download-artifact@bf251b5aa9c2f7eeb574a96ee720e24f801b7c11 # v6
+      uses: dawidd6/action-download-artifact@b6e2e70617bc3265edd6dab6c906732b2f1ae151 # v21
```
mcollins-ttd, line 74: **Claude Review:** `dawidd6/action-download-artifact` jumped fifteen major versions (v6 → v21). The inputs used here (`name`, `name_is_regexp`, `run_id`, `skip_unpack`) are all present in the current action schema, so the call site appears compatible. The v13 release converted the package from CommonJS to ESM, which is an internal change with no impact on callers. No action required, but a smoke-run of the AMI build workflow after merge is recommended given the size of the version gap.

.github/actions/build_ami/action.yaml
```diff
@@ -94,14 +94,14 @@ runs:
         ls ./scripts/aws/uid2-operator-ami/artifacts/ -al
 
     - name: Configure UID2 AWS credentials
-      uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a # v4
+      uses: aws-actions/configure-aws-credentials@e7f100cf4c008499ea8adda475de1042d6975c7b # v6.2.0
```
mcollins-ttd, line 97: **Claude Review:** `aws-actions/configure-aws-credentials` jumped two major versions (v4 → v6.2.0). The only confirmed breaking change in v5 was stricter boolean-input validation, and v6 requires runner `v2.327.1+` for Node 24 support. The inputs used here (`aws-region`, `role-to-assume`) are still valid in v6.2.0, so no change is needed — but worth confirming the self-hosted runners (if any) meet the v2.327.1+ runner version requirement before this merges.

.github/actions/build_ami/action.yaml
```diff
@@ -124,7 +124,7 @@ runs:
 
     - name: Setup Packer
       id: setup-packer
-      uses: hashicorp/setup-packer@c3d53c525d422944e50ee27b840746d6522b08de # main
+      uses: hashicorp/setup-packer@3286471d6cc6756d056a0b199fea5e0becdbc189 # v3.3.0
```
mcollins-ttd, line 127: **Claude Review:** Positive improvement — `hashicorp/setup-packer` was previously pinned to `main` (non-deterministic, mutable ref). The new pin to `v3.3.0` with a SHA is deterministic and consistent with repo convention.

mcollins-ttd (commented):
The upgrades are well-structured and the commit SHA pinning is correct throughout. The two significant concerns are the large major-version jumps (v4→v6.2.0 for configure-aws-credentials and v6→v21 for action-download-artifact) and the absence of a check for whether other workflow/action files in the repo also reference the old SHAs. The inputs used in the diff are confirmed valid in the current action schemas, and no breaking API changes were identified for the specific inputs in use.

sophia-chen-ttd (approved):

PR #2631, opened 2026-07-02
UID2-6767: publish a GitHub pre-release for scheduled and standalone operator builds

BehnamMozafari:
## Problem
`uid2-deployment` #4068 ([UID2-6767]) adds a pre-deployment **GitHub-Release-existence gate**: deploying a non-SNAPSHOT operator image version **hard-blocks** unless a published GitHub Release exists for that version (no override). But operator produces deployable image versions on build paths that publish **no** Release, so those deploys/rollbacks would block. Two such paths exist:

1. **Scheduled full-train** — `publish-all-operators.yaml` runs daily (`cron: 0 0 * * *`) and its `createRelease` job (which publishes a pre-release, per UID2-7340) was gated `if: github.event_name == 'workflow_dispatch'`. So scheduled builds pushed a `v<version>` tag but **no Release**. The frequently-deployed versions (e.g. `5.70.25`) come from here → tags exist up to `v5.70.193` with no matching Releases.
2. **Standalone public build** — `publish-public-operator-docker-image.yaml` (dispatched directly, incl. on `main`) hard-coded `force_release: 'no'`, so it produced a deployable image with **no tag and no Release**.

This PR makes every deployable operator version publish a pre-release automatically, so the gate passes with no manual intervention — while keeping exactly **one** Release per version.

## Changes

**1. Run `createRelease` on scheduled builds** (`publish-all-operators.yaml`)
```diff
-    if: github.event_name == 'workflow_dispatch'
+    if: github.event_name == 'workflow_dispatch' || github.event_name == 'schedule'
```
Reuses the existing pre-release logic. Scheduled runs always resolve `release_type` to `patch` (never `Snapshot`), so no Snapshot-release edge case.

**2. Alert on `createRelease` failure** (`publish-all-operators.yaml`)
`createRelease` now runs unattended daily, so it was added to `notifyFailure.needs` — a failure isolated to it (all build jobs green) would otherwise skip the Slack alert.

**3. `force_release` input on the public build** (`publish-public-operator-docker-image.yaml` + `publish-all-operators.yaml`)
Replaces the hard-coded `force_release: 'no'` with an input: default `'branch'` on `workflow_dispatch`, default `'no'` on `workflow_call`. The parent `buildPublic` passes `'no'` explicitly, so only the parent's `createRelease` cuts the single combined Release — **no double-up**. Behaviour:

| Trigger | Result |
|---|---|
| Standalone dispatch on `main`, Patch/Minor/Major | pre-release + tag + version bump |
| Standalone dispatch on a feature branch | no release (branch guard in `check_branch_and_release_type`) |
| Standalone dispatch, Snapshot (the default `release_type`) | no release |
| Called by `publish-all-operators` | no release; parent's `createRelease` owns it |

`release_type: Snapshot` + `force_release: yes` is rejected by a fail-fast guard in `check_major` — `check_branch_and_release_type` sets `is_release=true` unconditionally for `force_release: yes`, with no release-type check, so without the guard this combination would force a Snapshot Release/tag.

**4. Document the standalone side effect** (`publish-public-operator-docker-image.yaml`)
The `force_release` input description now states that a Patch/Minor/Major dispatch on `main` commits a version bump + pushes the `v<version>` tag and publishes a **public-operator-only** pre-release — so it's visible in the Run-workflow UI, not just git history.

**5. Serialize version bumps with a shared concurrency group** (`publish-public-operator-docker-image.yaml` + `publish-all-operators.yaml`)
Change 3 means a standalone Patch/Minor/Major dispatch on `main` now bumps the pom + pushes a tag — which can **race** the scheduled full-train doing the same (push conflict / duplicate-tag failure). Added a shared group `operator-version-bump-${{ github.ref }}` so the two can't bump at once:
- **job-level on the `image` job** of `publish-public-operator-docker-image.yaml` — scoped to just the version-bump/tag-push job, not the `e2e`/artifact-collection jobs that run after it.
- **job-level on the `start` job** of `publish-all-operators.yaml` — for the same reason, and so the lock frees before `buildPublic` (a reusable call sharing the group) runs, avoiding a deadlock.
- `cancel-in-progress: false` queues one in-flight bump/release rather than killing it. GitHub Actions only queues one waiter deep, so a third concurrent trigger would still cancel the queued second one — `notifyFailure` now also fires on `cancelled()` (not just `failure()`) as a backstop for that case.

## Why this is safe (verified in two review passes over the reusable-workflow call graph)
- **No Snapshot releases** — scheduled runs are `patch`; standalone `release_type` defaults to `Snapshot` → `is_release=false`; `Snapshot` + `force_release: yes` is explicitly rejected.
- **No double-release** — `force_release: 'no'` → `is_release=false` → every release step in `shared_create_releases` is a no-op; parent's `createRelease` is the sole releaser on the full train.
- **No deadlock** — job-level concurrency on both `start` and `image` frees before the next stage sharing the group runs.
- **Tag-before-verify ordering** holds — the tag is pushed by `commit_pr_and_merge` before `shared_create_releases`' "verify tag exists" step.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/publish-all-operators.yaml
+++ b/.github/workflows/publish-all-operators.yaml
@@ -268,7 +268,12 @@ jobs:
   notifyFailure:
       name: Notify Slack on Failure
       runs-on: ubuntu-latest
-      if: failure() && github.ref == 'refs/heads/main'
+      # cancelled() catches the case where a needed job (e.g. buildPublic's
+      # image job) loses the operator-version-bump-* concurrency group to a
+      # concurrent standalone dispatch and gets evicted — that surfaces as
+      # "cancelled", not "failure", and failure() alone would miss it, silently
+      # reintroducing the no-Release gap this workflow exists to close.
+      if: (failure() || cancelled()) && github.ref == 'refs/heads/main'
       needs: [start, buildPublic, buildGCP, buildAzure, buildAWS, buildAMI, createRelease]
       steps:
         - name: Send Slack Alert
--- a/.github/workflows/publish-all-operators.yaml
+++ b/.github/workflows/publish-all-operators.yaml
@@ -160,11 +160,9 @@ jobs:
     name: Create Release
     runs-on: ubuntu-latest
     # Also publish on scheduled builds, not just manual dispatch: the deployed
-    # operator versions come from the daily scheduled run, so gating this on
-    # workflow_dispatch left those versions tagged but with no GitHub Release.
-    # The uid2-deployment pre-deploy release gate (UID2-6767) then hard-blocks
-    # any deploy/rollback of such a version. Scheduled runs are always
-    # release_type 'patch' (never Snapshot), so no Snapshot-release edge case.
+    # operator versions may come from the daily scheduled run, so
+    # the uid2-deployment pre-deploy release gate won't hard-block
+    # any deploy/rollback of a version due to not having a release.
     if: github.event_name == 'workflow_dispatch' || github.event_name == 'schedule'
     needs: [start, buildPublic, buildGCP, buildAzure, buildAWS, buildAMI]
     steps:
--- a/.github/workflows/publish-all-operators.yaml
+++ b/.github/workflows/publish-all-operators.yaml
@@ -151,7 +151,13 @@ jobs:
   createRelease:
     name: Create Release
     runs-on: ubuntu-latest
-    if: github.event_name == 'workflow_dispatch'
+    # Also publish on scheduled builds, not just manual dispatch: the deployed
+    # operator versions come from the daily scheduled run, so gating this on
+    # workflow_dispatch left those versions tagged but with no GitHub Release.
+    # The uid2-deployment pre-deploy release gate (UID2-6767) then hard-blocks
+    # any deploy/rollback of such a version. Scheduled runs are always
+    # release_type 'patch' (never Snapshot), so no Snapshot-release edge case.
+    if: github.event_name == 'workflow_dispatch' || github.event_name == 'schedule'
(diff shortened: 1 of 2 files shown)


BehnamMozafari:
Added a 3rd commit (`5cf53b0c`) covering the **other** path to the same problem: standalone `workflow_dispatch` of `publish-public-operator-docker-image.yaml`.

That workflow hard-coded `force_release: 'no'`, so a manual build on `main` produced a deployable image with **no tag and no Release** — the gate (UID2-6767) would block deploying it. It's dispatched on `main` regularly (the parent-invoked calls don't show in run history since reusable calls run inside the caller).

Change: a `force_release` input — `'branch'` default on `workflow_dispatch`, `'no'` default on `workflow_call` — passed through to the shared build. `publish-all-operators` now passes `'no'` explicitly, so only its `createRelease` job cuts the single combined Release (**no double-up**).

Behavior matrix:
| Trigger | `force_release` | Result |
|---|---|---|
| Standalone dispatch on `main`, Patch/Minor/Major | `branch` | ✅ pre-release + tag + version bump |
| Standalone dispatch on a feature branch | `branch` | no release (branch guard) — unchanged for test builds |
| Standalone dispatch, Snapshot | `branch` | no release |
| Called by `publish-all-operators` | `no` | no release; parent's `createRelease` owns it |

⚠️ **Reviewer sign-off needed:** this changes standalone-dispatch behavior on `main` — such a build now commits the version bump + pushes the `v<version>` tag (same mechanism as the parent's `start` job) instead of just building an image. Feature-branch/Snapshot dispatches are unaffected. Flag if you'd prefer the dispatch default to be `'no'` (opt-in) instead.

jon8787 (approved):

PR #2639, opened 2026-07-07
UID2-7426: add check_jira_key required-status-check caller

jon8787:
Adds the check_jira_key required-status-check caller (UID2-7426). Non-blocking until the terraform required_status_checks rule is flipped on. Identical to the reviewed uid2-attestation-api#49.

### DIFF
--- a/.github/workflows/check-jira-key.yaml
+++ b/.github/workflows/check-jira-key.yaml
@@ -1,0 +1,16 @@
+name: Check Jira Key
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


BehnamMozafari (approved):
Verified byte-identical to the approved UID2-7426 check_jira_key caller baseline (single .github/workflows/check-jira-key.yaml, +16/-0, job check_jira_key, action @v3). Approving as part of the rollout batch.

PR #2696, opened 2026-08-17
UID2-7080: Add SLSA provenance for private operator artifacts

swibi-ttd:
## Summary
- Add producer-side SLSA attest+verify for private operator images (GCP GHCR/GAR, Azure, EKS), EIFs/PCR0, AMI metadata, enclave-id manifests, and release ZIP archives.
- Gate non-snapshot tag promotion and release upload on successful attestation verification (`attest_file`, shared `attest_image@v3`, plus registry `--bundle-from-oci` checks).
- Document `gh attestation verify` usage with `--signer-workflow` / `--bundle-from-oci`; snapshots continue to skip attestation.

## Notes
- Same-run version-bump commit identity gap is unchanged (accepted; shared with public image path / UID2-6764).
- No changes to `uid2-private-operator-release`, `uid2-deployment`, or consumer enforcement.

## Test plan
- [ ] `actionlint` / workflow YAML review of changed publish paths
- [ ] Dry-run or snapshot publish confirms attestation steps are skipped for `-SNAPSHOT`
- [ ] Non-snapshot GCP publish: digest push → attest GHCR+GAR → `--bundle-from-oci` verify → tag promotion; both tags resolve to attested digest
- [x] Non-snapshot Azure / EKS publish: same digest-first + attest + verify + promote flow
- [ ] AWS Nitro: EIF/PCR0 and standalone deployment ZIPs attest before upload/release
- [ ] AMI build verifies consumed EIF provenance before Packer; AMI metadata attested before artifact upload
- [ ] `publish-all-operators` attests combined release ZIPs before `action-gh-release`
- [x] Spot-check `gh attestation verify` commands from README against a real non-snapshot artifact set

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/actions/build_ami/action.yaml
+++ b/.github/actions/build_ami/action.yaml
@@ -135,10 +135,10 @@ runs:
           exit 1
         fi
         # EIF may have been attested by the standalone Nitro workflow or by
-        # Publish All Operators calling that workflow via workflow_call.
+        # the same reusable Nitro workflow invoked by Publish All Operators.
         gh attestation verify "$EIF_PATH" \
           --repo "$EIF_REPO" \
-          --cert-identity-regex "^https://github\\.com/${EIF_REPO}/\\.github/workflows/(publish-aws-nitro-eif|publish-all-operators)\\.ya?ml@"
+          --signer-workflow "${EIF_REPO}/.github/workflows/publish-aws-nitro-eif.yaml"
 
     - name: Prepare EIF archive for Packer
       shell: bash
--- a/.github/actions/build_ami/action.yaml
+++ b/.github/actions/build_ami/action.yaml
@@ -42,6 +42,9 @@ outputs:
   enclave_id:
     description: the Enclave_ID of the EIF used to create the AMI
     value: ${{ steps.extractAmiIdAndEnclaveId.outputs.ENCLAVE_ID }}
+  measurement_path:
+    description: Path to the generated AMI measurement metadata file
+    value: ${{ steps.measurementPath.outputs.path }}
 
 runs:
   using: "composite"
@@ -87,10 +90,6 @@ runs:
         FILE=$(echo $ARTIFACTS | jq -r '.[0].name')
         unzip -o -d ./scripts/aws/uid2-operator-ami/artifacts $FILE.zip
         rm $FILE.zip
-        cd "./scripts/aws/uid2-operator-ami/artifacts/"
-        zip "uid2operatoreif.zip" "uid2operator.eif"
-        cd -
-        rm ./scripts/aws/uid2-operator-ami/artifacts/uid2operator.eif
         ls ./scripts/aws/uid2-operator-ami/artifacts/ -al
 
     - name: Configure UID2 AWS credentials
@@ -121,6 +120,34 @@ runs:
         VERSION_NUMBER=$(cat ./artifacts/version_number.txt)
         echo "VERSION_NUMBER=$VERSION_NUMBER" >> $GITHUB_OUTPUT
         echo $VERSION_NUMBER
+
+    - name: Verify consumed EIF provenance
+      if: ${{ !contains(steps.versionNumber.outputs.VERSION_NUMBER, 'SNAPSHOT') }}
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github_token }}
+        EIF_PATH: ${{ github.workspace }}/scripts/aws/uid2-operator-ami/artifacts/uid2operator.eif
+        EIF_REPO: ${{ inputs.eif_repo_owner }}/${{ inputs.eif_repo_name }}
+      run: |
+        set -euo pipefail
+        if [[ ! -f "$EIF_PATH" ]]; then
+          echo "::error file=${EIF_PATH}::Expected EIF subject is missing before AMI build"
+          exit 1
+        fi
+        # EIF may have been attested by the standalone Nitro workflow or by
+        # Publish All Operators calling that workflow via workflow_call.
+        gh attestation verify "$EIF_PATH" \
+          --repo "$EIF_REPO" \
+          --cert-identity-regex "^https://github\\.com/${EIF_REPO}/\\.github/workflows/(publish-aws-nitro-eif|publish-all-operators)\\.ya?ml@"
+
+    - name: Prepare EIF archive for Packer
+      shell: bash
+      run: |
+        set -euo pipefail
+        cd "./scripts/aws/uid2-operator-ami/artifacts/"
+        zip "uid2operatoreif.zip" "uid2operator.eif"
+        rm "uid2operator.eif"
+        ls -al
 
     - name: Setup Packer
       id: setup-packer
@@ -189,8 +216,10 @@ runs:
         cat euid_AMI_measurement.txt
         ls -al
 
-    - name: Upload artifacts
-      uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
-      with:
-        name: ${{ inputs.identity_scope }}_AMI_measurement
-        path: ./scripts/aws/uid2-operator-ami/${{ inputs.identity_scope }}_AMI_measurement.txt
+    - name: Set measurement metadata path
+      id: measurementPath
+      shell: bash
+      env:
+        IDENTITY_SCOPE: ${{ inputs.identity_scope }}
+      run: |
+        echo "path=${GITHUB_WORKSPACE}/scripts/aws/uid2-operator-ami/${IDENTITY_SCOPE}_AMI_measurement.txt" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/publish-all-operators.yaml
+++ b/.github/workflows/publish-all-operators.yaml
@@ -159,6 +159,11 @@ jobs:
   createRelease:
     name: Create Release
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
+      attestations: write
+      artifact-metadata: write
     # Also publish on scheduled builds, not just manual dispatch: the deployed
     # operator versions may come from the daily scheduled run, so
     # the uid2-deployment pre-deploy release gate won't hard-block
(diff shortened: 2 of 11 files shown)


UID2SourceAdmin:
## 🤖 AI Code Review

_Generated by Azure OpenAI (gpt-5). Automated — human review is still required._

## Summary
This PR adds artifact/image provenance attestation across the release pipelines: local file attestations for EIFs, manifests, AMI measurement files, and release archives; OCI attestation/verification for published images; digest-first image publishing with post-promotion digest checks; and updated README guidance for consumers to verify provenance.

## Strengths
- Verifying the consumed EIF before the AMI build is a strong supply-chain hardening step.
- The digest-first push flow plus post-tag-resolution check is a good pattern: it avoids attesting mutable tags and then confirms the promoted tags still resolve to the attested digest.
- Permissions were tightened explicitly on the jobs that mint attestations, instead of relying on broad defaults.

## Findings

### 🟡 Important

**Unpinned external attestation action reintroduces supply-chain risk**  
The new image attestation steps use `IABTechLab/uid2-shared-actions/actions/attest_image@v3` by mutable tag, while the rest of the workflows are SHA-pinned. That makes the provenance path itself depend on a retargetable reference. Pin this action to a full commit SHA and keep the version in a comment.  
`.github/actions/build_eks_docker_image/action.yaml:162`, `.github/workflows/publish-azure-cc-enclave-docker.yaml:183`, `.github/workflows/publish-gcp-oidc-enclave-docker.yaml:188`

**Fresh attestation verification has no retry/backoff, so these release paths may flake**  
Both new composite actions do a single immediate `gh attestation verify` right after the attestation is created/published. GitHub attestation APIs and OCI bundle publication are not always instantly visible, so transient `no attestations found` failures are likely here. Wrap verification in a bounded retry loop with small backoff.  
`.github/actions/attest_file/action.yaml:24-38`, `.github/actions/verify_oci_attestation/action.yaml:20-34`

### 🟢 Minor

**`verify_oci_attestation` is advertised as generic, but hardcodes the current repo**  
The action accepts any `signer_workflow`, but always verifies with `--repo "$GITHUB_REPOSITORY"`. That makes the abstraction incorrect for cross-repo signers and weaker than its input contract suggests. Add a `repo` input or derive `owner/repo` from `signer_workflow`.  
`.github/actions/verify_oci_attestation/action.yaml:29-34`

**README example weakens verification compared with the guidance above it**  
The new docs say to prefer `--signer-workflow`, but the public image example still uses `--signer-repo`, which would trust any workflow in `uid2-shared-actions`. Update the example to pin the exact shared workflow path as well.  
`README.md:83-92`

## Verdict
**Ready to merge?** `With fixes`

**Reasoning:** The overall direction is solid and materially improves provenance coverage, but the unpinned attestation action and lack of retry/backoff in the new verification steps are both worth fixing before relying on this in release pipelines.

swibi-ttd:
^ addressed initial comments on AI review but looks like it didn't work when i re-triggered the review workflow

BehnamMozafari:
Can you see if you can run a smoketest to check the new actions or one of the workflows, usually helps check the job permissions and signer repo/wf

BehnamMozafari (commented):
Reviewed the full diff plus the pinned upstream actions (`actions/attest@v4.1.0`, `docker/build-push-action@v7.2.0`, `buildx imagetools create`), `uid2-shared-actions/attest_image@v3`, and `gh attestation verify` flag semantics. Four comments inline, roughly in severity order.

A few things I specifically checked and found **correct**, noting them so they don't get re-raised: `imagetools create` preserves the attested digest for a single index source (`util/imagetools/create.go`, the `len(srcs) == 1 && len(ann) == 0` carbon-copy branch); the quoted multi-registry `outputs:` syntax survives `build-push-action`'s CSV parsing (`{ignoreComma: true, quote: false}`); `actions/attest@59d8942` is genuinely v4.1.0 and defaults to SLSA provenance with no predicate input; and `--repo` + `--signer-workflow` are compatible, with the prefix match on `job_workflow_ref` correctly keeping the reusable workflow as signer. Two real bug fixes in here as well — the new `Zip deployment artifacts` step finally creates the archives `Create release` was already referencing, and `Prepare EIF archive for Packer` now runs on the `operator_release` path where `uid2operatoreif.zip` was previously never created.

.github/workflows/publish-aws-nitro-eif.yaml
```diff
@@ -150,7 +178,15 @@ jobs:
     name: Cleanup Building AWS Image
     runs-on: ubuntu-latest
     needs: [start, buildUID2EIF, buildEUIDEIF]
+    permissions:
+      contents: read
```
BehnamMozafari: **`contents: read` will break the release step in this same job.**

This job had no `permissions:` block before, so it inherited the repository default — which is `write` (`gh api repos/IABTechLab/uid2-operator/actions/permissions/workflow` returns `"default_workflow_permissions": "write"`). Pinning `contents: read` narrows it, but `Create release` (line 261, `softprops/action-gh-release`) still runs in this same job.

Failure path: a standalone `Publish AWS Nitro EIFs` dispatch with `release_type: Patch/Minor/Major` and an empty `version_number_input` will build both EIFs, attest them, zip and attest the deployment archives, and then fail with HTTP 403 on `POST /repos/.../releases`.

The `publish-all-operators` path won't catch this, because it passes a non-empty `version_number_input`, which makes the `if` on the release step false. So it only surfaces on the manual/standalone dispatch.

Suggest `contents: write`.

.github/workflows/publish-gcp-oidc-enclave-docker.yaml
```diff
@@ -100,28 +102,8 @@ jobs:
...
-          service_account: ${{ vars.GCP_PUBLISH_SERVICE_ACCOUNT }}
-          access_token_lifetime: 300s
-
-      - name: Log in to the GCP Registry
-        uses: docker/login-action@650006c6eb7dba73a995cc03b0b2d7f5ca915bee # v4.2.0
-        with:
-          registry: ${{ env.GCP_REGISTRY }}
-          username: oauth2accesstoken
-          password: ${{ steps.gcp_auth.outputs.access_token }}
+      - name: Set up Docker Buildx
```
BehnamMozafari, line 105: **Adding the buildx container driver changes the published private image from a plain manifest to an OCI image index — which changes the digest that feeds the GCP enclave ID.**

Same pattern at `publish-azure-cc-enclave-docker.yaml:107` and `.github/actions/build_eks_docker_image/action.yaml:114`.

Without `setup-buildx-action` these jobs used the default `docker` driver. Checked against the live registry as it stands today:

- `ghcr.io/iabtechlab/uid2-operator:5.70.259-gcp-oidc` → `application/vnd.docker.distribution.manifest.v2+json` (plain manifest)
- `ghcr.io/iabtechlab/uid2-operator:5.70.259` (public image, already on buildx) → `application/vnd.oci.image.index.v1+json`, with an `attestation-manifest` attached

On the container driver, `docker/build-push-action` v7.2.0 adds `--attest type=provenance,mode=max` for public repos on any non-`load` output (`src/context.ts:254-265`), and attaching that attestation requires the pushed artifact to become an index. So `Push ... by digest` will now publish an index, and `steps.push-to-docker.outputs.digest` becomes an **index** digest rather than a manifest digest.

That value is hashed directly into the GCP enclave ID by `scripts/gcp-oidc/generate-deployment-artifacts.sh` (`echo -n "V1,false,$IMAGE_DIGEST" | sha256 | base64`) — the ID partners register and that Confidential Space must reproduce at runtime. Reading the CS launcher (`google/go-tpm-tools/launcher/container_runner.go:146`, `image.Target().Digest`) suggests containerd resolves a tag to the index digest, so it *should* stay consistent — but this is an unstated change to the one value that gates every GCP private operator's attestation, and `az confcom acipolicygen` is now being handed an index too.

Test-plan items 3 and 4 (end-to-end GCP CS / Azure CC attestation) cover exactly this and are still unchecked — worth running them before merge. If the assumption doesn't hold, GCP private operators fail attestation after the release ships, in partner environments.

.github/actions/build_ami/action.yaml
```diff
@@ -122,6 +121,34 @@ runs:
         echo "VERSION_NUMBER=$VERSION_NUMBER" >> $GITHUB_OUTPUT
         echo $VERSION_NUMBER
 
+    - name: Verify consumed EIF provenance
```
BehnamMozafari, line 124: **This gate blocks AMI rebuilds of any already-shipped version.**

`Verify consumed EIF provenance` runs for every non-SNAPSHOT build and hard-fails if `gh attestation verify` finds nothing. EIFs built before this PR have no attestation on record, so dispatching `Build UID2 and EUID AMIs` with `operator_release: 5.70.259` — or an `operator_run_number` from any pre-merge run — now aborts before Packer starts.

That removes the ability to rebuild or repair an AMI for a version that is already out: Marketplace republish, region copy, replacing a deleted AMI, or an ansible/Packer-level fix that doesn't touch the enclave measurements. SNAPSHOT builds are exempt, so day-to-day development won't surface it; the first person to hit it will be whoever needs an old-version rebuild.

A version floor in the check (or an explicit documented bypass input) would be the robust fix, but **at minimum please document the cut-over version in the runbook**, so that when someone does hit this the failure is immediately recognisable rather than looking like a broken build.

.github/workflows/publish-all-operators.yaml
```diff
@@ -159,6 +159,11 @@ jobs:
   createRelease:
     name: Create Release
     runs-on: ubuntu-latest
+    permissions:
```
BehnamMozafari, line 162: **This block narrows the token for the changelog step, dropping `pull-requests`.**

This job previously inherited the repository default (`write` across the board). `mikepenz/release-changelog-builder-action` (line 228) enumerates merged PRs for `pr_template`. Public-repo reads generally still succeed with a reduced token, so this is unlikely to hard-fail — but the narrowing is a side effect of adding the attestation permissions rather than an intended change.

Suggest adding `pull-requests: read` here, and the same for the changelog step in `publish-aws-nitro-eif.yaml`.

.github/workflows/publish-aws-nitro-eif.yaml
swibi-ttd: good catch, updated to `write`

.github/workflows/publish-all-operators.yaml
swibi-ttd: fixed as suggested

.github/actions/build_ami/action.yaml
swibi-ttd: added note in README

.github/workflows/publish-gcp-oidc-enclave-docker.yaml
swibi-ttd: updated GCP, Azure, EKS push steps with `provenance: false` (but still keeps the separate SLSA attestations)

swibi-ttd:
> Can you see if you can run a smoketest to check the new actions or one of the workflows, usually helps check the job permissions and signer repo/wf

Ran Azure CC and GCP pipelines from this branch as tests and tested out the attestation (more details in the ticket comments)

BehnamMozafari (approved):

PR #2718, opened 2026-09-02
UID2-7641: Remove support for V2 refresh tokens

samin-ttd:
### Summary
V2 refresh tokens are AES-CBC encrypted, which is what makes this class of padding oracle possible in the first place. All refresh tokens issued today are V3 (AES-GCM, an authenticated mode with no padding-oracle equivalent), so V2 refresh token decoding is dead weight kept only for backward compatibility. Removing it eliminates the vulnerable code path entirely rather than just papering over its error handling.

### Change
- Added a catch-all `Exception` handler around refresh token decoding.
- Removed `decodeRefreshTokenV2` and the V2 branch in `decodeRefreshToken` — any V2-formatted refresh token now fails the same generic version check as other malformed input.
- Removed the now-unreachable `encodeV2(RefreshToken, ...)` and its `case V2` in `encode(RefreshToken, ...)` (refresh tokens have only ever been issued as V3 since `UIDOperatorService.refreshTokenVersion` is hardcoded to `TokenVersion.V3`).

### Testing
- [x] Send requests crafted to trigger padding against the endpoint both before and after the change: ensure the response is 500 before and 400 after.
- [x] Full test suite passes (`mvn test`): 761 tests, 0 failures, 0 errors.
- [x] Confirmed no remaining references to `decodeRefreshTokenV2` / `encodeV2(RefreshToken...)` anywhere in the codebase.

### DIFF
--- a/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
+++ b/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
@@ -1525,15 +1525,7 @@ public class UIDOperatorVerticle extends AbstractVerticle {
         } catch (ClientInputValidationException cie) {
             LOGGER.warn("Failed to decode refresh token for site ID: " + rc.data().get(Const.RoutingContextData.SiteId), cie);
             return RefreshResponse.Invalid;
-        } catch (Exception e) {
-            if (e.getCause() instanceof BadPaddingException || e.getCause() instanceof IllegalBlockSizeException) {
-                LOGGER.warn("Failed to decode refresh token for site ID: " + rc.data().get(Const.RoutingContextData.SiteId), e);
-            } else {
-                LOGGER.error("Unexpected error decoding refresh token for site ID: " + rc.data().get(Const.RoutingContextData.SiteId), e);
-            }
-            return RefreshResponse.Invalid;
         }
-
         if (refreshToken == null) {
             return RefreshResponse.Invalid;
         }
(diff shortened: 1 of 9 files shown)


src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java
```diff
@@ -1525,7 +1525,15 @@ private RefreshResponse refreshIdentity(RoutingContext rc, String tokenStr) {
         } catch (ClientInputValidationException cie) {
             LOGGER.warn("Failed to decode refresh token for site ID: " + rc.data().get(Const.RoutingContextData.SiteId), cie);
             return RefreshResponse.Invalid;
+        } catch (Exception e) {
+            if (e.getCause() instanceof BadPaddingException || e.getCause() instanceof IllegalBlockSizeException) {
+                LOGGER.warn("Failed to decode refresh token for site ID: " + rc.data().get(Const.RoutingContextData.SiteId), e);
+            } else {
+                LOGGER.error("Unexpected error decoding refresh token for site ID: " + rc.data().get(Const.RoutingContextData.SiteId), e);
+            }
+            return RefreshResponse.Invalid;
```
mcollins-ttd: I'm not sure we need/want this now that the AES-CBC path is gone. A bug in our code would look like bad client input and we'd lose alerting on it.

samin-ttd: Agreed reverted it.

mcollins-ttd (dismissed):

mcollins-ttd (approved):

